PR #5, opened 2019-06-11
Feature/process block

aabounegm:


components/landing/process-step.vue
```diff
@@ -0,0 +1,38 @@
+<template functional>
+  <div class="container">
+    <img :src="props.img" />
+    <div class="description">
+      <h3 class="title">{{ props.title }}</h3>
+      <p class="subtitle">{{ props.subtitle }}</p>
+      <nuxt-link
```
VanishMax: If assigned - should be <a>, not <nuxt-link>. Try clicking on the link - now it redirects relatively

components/landing/process-step.vue
```diff
@@ -0,0 +1,38 @@
...
+
+<script>
+  export default {
+    props: {
+      img: String,
+      title: String,
+      subtitle: String,
+      buttonLink: {
+        type: String,
+        default: '#',
```
VanishMax: Make default be `'/'` cause we're not in plain html and can do this

constants/process.js
```diff
@@ -0,0 +1,31 @@
+export default [
+  {
+    title: 'Volunteering',
+    subtitle: 'Innopolis University needs heroes like you to organize world-class events!',
+    img: 'images/landing/volunteering.png',
+    buttonLink: '#',
```
VanishMax: And when it is a default link - no need to specify

aabounegm: I put it there to make it easier to change when we put the actual link.
You will see it right in front of you without having to go back to the component and check the prop name and so on

components/landing/process-step.vue
aabounegm: But since we are building an SPA, doesn't it make sense to use `<nuxt-link>` which extends `<router-link>` so that links don't open in a new tab?
(I'll fix it for the external link, of course)

components/landing/process-step.vue
aabounegm: I don't understand. Why isn't `'#'` valid? It does the job of not making the link go anywhere

VanishMax: You are right, but why do weed this hash in the url?

components/landing/process-step.vue
VanishMax: Only for external, of course

components/landing/process-step.vue
aabounegm: because we don't want it to redirect anywhere. making it `/` will cause is to redirect to the root. For example, assume our site got hosted at "https://my.university.innopolis.ru/innopoints"; you would want the button (if not assigned a URL, which shouldn't be the case anyways!) to keep it at the same page, not go back to the dashboard

illright:
@aabounegm please, correct your commit names as per the guideline

aabounegm:
What's wrong with the merging? You both approved, why is the "merge" button disabled?

illright:
Max's review is now dismissed due to the rebase, he has to do it again

PR #6, opened 2019-06-11
Feature/store block

aabounegm:


components/landing/store-card.vue
```diff
@@ -0,0 +1,37 @@
+<template functional>
+  <div class="card shadow-2">
+    <img :src="props.img" :style="{background: props.backgroundColor}" />
```
illright: About that: the database will store a hex color code for each item in the store, which will be the `color` in the props. The `backgroundColor` is supposedly obtained from lightening the `color` to a certain extent, but the whole process of that isn't streamlined yet, so no need to worry about that. What would be good though is to allow for that change to happen seamlessly, perhaps wrap the background value in some function that will, for now, return a fixed mapping, and in the future will lighten the color

aabounegm: Can you elaborate more on the 

> lightening the `color` to a certain extent

?

PR #13, opened 2019-06-22
Feature/events

aabounegm:
First 2 tasks (copy markup and sample data).
Switching between user and admin view is on hold until Vuex is integrated in the project.


components/events/event.vue
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+      </div>
+      <a :href="url" class="btn filled green shadow-1">see details</a>
+    </div>
+  </div>
+</template>
+
+<script>
+  import Vue from 'vue';
+  export default Vue.extend({
```
VanishMax: just `export default {}`

constants/events/events.js
```diff
@@ -0,0 +1,41 @@
+export default [
+  {
+    img: '/images/browser/chess-club.png',
```
VanishMax: Change static directory to _events_

components/events/event.vue
aabounegm: Both are effectively equivalent, but this one helps my IDE (VS Code) with typings (even if it is not Typescript) which I prefer.

components/events/event.vue
```diff
@@ -3,6 +3,15 @@
     <img :src="img" />
     <h2>{{ title }}</h2>
     <div class="content">
+      <div class="compact" v-if="$store.state.auth.isAuth">
```
VanishMax: Bad practice. It's much better to move **isAuth** into _computed_ using mapState

components/events/event.vue
VanishMax: IDE is changable, but code should look pretty. This type requires one extra useless import

illright: Hate to intervene when I don't really know sh!t about what's going on, but typings don't hurt code quality. Excess imports do, however. Any way to preserve the typings by just exporting the object with the `props` property, without throwing it on top of the Vue object?

containers/events/events.vue
```diff
@@ -0,0 +1,29 @@
+<template>
```
illright: This file feels awfully excessive. As well as the `EventsTagline` component. Do we really need to make everything a component (which potentially introduces excess `div` s in the markup, which is **not acceptable**)? IMHO, the only thing that should really be a component is something that is data-dependent (event cards) or has logic integrated (filters block). The header and tagline are not deserving of a component to me. And if that helps fight excess markup, by all means, go for it

components/events/event.vue
aabounegm: @illright, by typings I didn't mean the props or the component's interface, but rather while developing getting helpful auto-complete options. Also, the imports all share the same object, so no extra load is added to the build.
However, if you don't like it I'll remove it.

containers/events/events.vue
aabounegm: Components do not introduce an extra `div`. They get replaced by the template's root element. They help with abstraction and avoiding repetition. Sometimes their main goal is to simply isolate a block that has an independent purpose from an otherwise very lengthy code.

components/events/event.vue
illright: @aabounegm yeah, that's what I meant. I see that most type information is stored in the object you're extending `Vue` with, so any way to preserve that information but without the Vue object?
For a second thought, if you say that all imports share the same object (which is kind of strange, but could be the case), then I'm ok with the import and the typing

containers/events/events.vue
illright: Yeah, it's just that sometimes a component may not have a root element, which forces to create container elements which weren't part of the markup. It's not a hit at this particular code piece, I'd just like to express that if we are picking between clean markup and short files, we should probably go for clean markup

components/events/event.vue
aabounegm: Without actually using Typescript (which I voted for in the beginning), there is no other way. VS Code does provide some hints even with `export default {};`, though not as good. But if this is the style we agreed on, then I don't mind.
As for the import thing, all imports do indeed share the same object. This is the case with all Javascript imports (be it `require` in Node, ES6 imports, CommonJS libraries, ...)

containers/events/events.vue
aabounegm: You mean not have a template at all? I don't think a component like that would be useful. In case you meant multiple root elements, this is not even allowed in Vue. Templates can only have one root element.

illright: What I mean is, consider, for example, a component that consists of two sibling elements. The only way to make it a Vue component is to introduce an outer tag to wrap those two

PR #14, opened 2019-06-26
Feature/pagination

aabounegm:


aabounegm:
I don't know why there are so many commits; should be just the last 5. Better check the "Files Changed" tab

components/pagination.vue
```diff
@@ -0,0 +1,217 @@
+<!-- Adapted from https://github.com/lokyoung/vuejs-paginate -->
+<template>
+  <ul :class="containerClass">
+    <li v-for="page in pages" :class="[pageClass, page.selected ? activeClass : '', page.disabled ? disabledClass : '']" :key="page.index">
+      <template v-if="page.breakView">
+        <input 
+          v-if="editing === page.index"
+          :class="pageLinkClass"
+          @keyup.enter.prevent="handlePageSelected(parseInt($event.target.value, 10));"
```
VanishMax: Brrrrrr.. semicolon? :)

components/pagination.vue
```diff
@@ -0,0 +1,217 @@
...
+        innerValue: 1,
+        containerClass: 'pages',
+        pageClass: 'page-item',
+        pageLinkClass: 'page-link-item',
+        editing: -1,
+      };
+    },
+    computed: {
+      selected: {
+        get() {
```
VanishMax: These functions are not used. Do we need them?

components/pagination.vue
```diff
@@ -0,0 +1,217 @@
...
+          }
+
+          // 3rd - loop thru high end of margin pages
+          for (let i = this.pageCount - 1; i >= this.pageCount - this.marginPages; i--) {
+            setPageItem(i);
+          }
+        }
+        return items;
+      },
+      firstPageSelected() {
```
VanishMax: Also two unused functions. What for?

components/pagination.vue
aabounegm: They are getters and setters for the property (https://www.w3schools.com/js/js_object_accessors.asp)
They are used everywhere you see `this.selected`.

aabounegm:
> 1. Design small fixes image

Are you possibly looking at the tablet/mobile version? I copied all the font sizes/weights/... from Figma

illright:
image
When the input is active, the hover effect of buttons should be disabled.

illright:
The font in the inputs is huge for some reason

illright:
Also, when there are only 3 pages, the spacing between them is way larger. Make it consistent.

illright:
Aaalso, don't display the pagination when there's only one page. And we should probably double, if not triple, the paging constant, as nobody likes clicking through pages. Let's double it for now

aabounegm:
I simply replaced the styles I added with the CSS file from the markup, so I assumed that this all was intended behaviour.

PR #29, opened 2019-07-30
Feature/card-grid

dark0n3:


containers/landing/store.vue
```diff
@@ -8,10 +8,11 @@
     </div>
 
     <div class="cards">
-      <Card
-        v-for="(item, i) in items"
-        :key="i"
+      <StoreCard
+        v-for="item in items.slice(0, 3)"
```
illright: Don't slice here, assume that the `items` array will be 3 in length. The API will have a limiting parameter which will provide that

pages/shop.vue
```diff
@@ -0,0 +1,41 @@
...
+    head: {
+      title: 'InnoStore',
+    },
+    components: {
+      StoreCard,
+      ShopTagline,
+    },
+    data() {
+      return {
+        items,
```
illright: Instead, slice here for now

static/css/shop/_filters-spec.scss
```diff
@@ -0,0 +1,139 @@
...
+
+#chk-000000 > input + .icon {
+  background: #000;
+  border-color: #000;
+  --r: 0;
+  --g: 0;
+  --b: 0;
+}
+
+#chk-a7a7a7 > input + .icon {
```
illright: That's hardly maintainable. Those IDs were meant for demo, they shouldn't end up in the actual frontend

components/ui/radio-group.vue
```diff
@@ -63,16 +63,27 @@
         this.$emit('input', item);
       },
       style(item) {
-        if(!item.color) {
-          return '';
+        if(!item.color) return '';
+
+        // Parse HEX colors
+        let outline = '';
+        let match = item.color.match(/#([0-9A-F]{2})([0-9A-F]{2})([0-9A-F]{2})/);
```
aabounegm: Probably need to match lowercase too, or match with `item.color.toUpperCase()` instead

constants/store.js
```diff
@@ -0,0 +1,126 @@
+export default [
```
aabounegm: Why is this file not inside a folder like the other pages?
And why is it named "store" here, but "shop" in other places?

static/css/shop/_main.scss
```diff
@@ -0,0 +1,66 @@
...
+.card {
+  max-width: 18em;
+  margin-top: 1em;
+}
+
+.card > .image {
+  position: relative;
+}
+
+.card > .image > img {
```
aabounegm: Try to use SCSS features for better readability.
```suggestion
.card {
  max-width: 18em;
  margin-top: 1em;
  >.image {
    position: relative;
    >img {
      width: 100%;
    }
  }
}
```
However, I wouldn't block the merge because of this. Just note for the future.

components/ui/radio-group.vue
VanishMax: That's right

constants/store.js
VanishMax: Shop has only one constant file while other pages have at least two. But naming is definitely wrong

PR #30, opened 2019-07-31
Feature/product edit

aabounegm:
Completes https://trello.com/c/PRYU7hNx/53-product-edit-page-color-picker-and-the-dropzone-with-dragdrop
Color picker and image uploader


components/ui/dropdown.vue
```diff
@@ -1,16 +1,33 @@
 <template>
-  <div :class="isOpen && 'open'" class="dropdown-shell">
+  <div :class="{open: isOpen}" class="dropdown-shell">
```
VanishMax: Let's wrap classes in single quotes. On the line 14 you put it, so do it everywhere for the sake of code consistency

components/ui/dropdown.vue
```diff
@@ -1,16 +1,33 @@
...
-    <div :class="right && 'right-edge'" class="dropdown">
-      <slot />
+    <div :class="{'right-edge': right}" class="dropdown">
+      <div class="relative-wrapper">
+        <button @click="isOpen = false" class="btn normal round close">
+          <img src="/images/ui-kit/x.svg" />
+        </button>
+        <div
+          v-for="(option, i) in options"
+          :key="option.value"
```
VanishMax: Dropdown is not only for the lists. There are could be just anything. Slot would work better

components/ui/dropdown.vue
aabounegm: The point is that it is a key in an object, which means that it needs to be wrapped in quotes when it contains any special characters (like 'right-edge' in line 14) but not otherwise

components/ui/dropdown.vue
aabounegm: That's why there is a slot inside it to customize how the item looks like. This provides a default interface to simplify the general case

aabounegm: Also, making the entire dropdown content a slot would seem like taking too much repeated work outside of the component, like the dropdown saying "it's not my job" ))

PR #31, opened 2019-08-01
Single product page

VanishMax:


containers/shop/balance.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+  <div class="balance-tooltip">
+    You have 1337
```
aabounegm: How about this number is moved to the store and a getter is used instead?

store/product.js
```diff
@@ -0,0 +1,27 @@
+import items from '../constants/shop';
+
+export default {
+  state() {
```
aabounegm: VueX has to do with data that is shared between components. Here, this information is even already depicted in the route, so it does not need to be shared or even stored.

pages/shop/index.vue
```diff
@@ -0,0 +1,54 @@
...
+          v-bind="item"
+        />
+      </div>
+    </section>
+    <p class="link-bottom padded">
+      Lacking innopoints?
+      <nuxt-link to="/events">Volunteer on events</nuxt-link> to fill in the shortage!
+    </p>
+
+    <ProductModal v-if="open" />
```
aabounegm: Why not use a nested route-view instead?

containers/shop/product-modal.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <div :class="{'modal-open': open}" class="modal-overlay column">
+    <BalanceTooltip />
+    <ProductCard v-bind="product" closable />
```
aabounegm: Can it close on clicking outside?
Also, the close button can hide behind the header, and would even be unclickable if the screen height is shorter than product modal
image

containers/shop/product-card.vue
```diff
@@ -0,0 +1,119 @@
...
+  <div class="card product with-image">
+    <Button
+      v-if="closable"
+      @click="closeModal"
+      normal
+      round
+      class="close"
+      img="/images/x.svg"
+    />
+    <Carousel :images="activeVariety.images" />
```
aabounegm: When a carousel is on the 2nd picture, then a variety (color) with only one picture is chose, the following happens:
image
This is because the "current" variable inside the carousel stays on 1 and doesn't reset to 0
Suggestion: add a watcher inside the carousel component to reset to 0 if the `images` prop changes
Suggestion 2: expose the "current" variable as a prop (maybe even `v-model`)

containers/shop/product-card.vue
```diff
@@ -0,0 +1,119 @@
...
+
+        <div class="labeled text">
+          <span class="label">Sizes</span>
+          <div role="group" class="radio-options">
+            <Chip
+              v-for="(val, key) in activeVariety.sizes"
+              :disabled="val === 0"
+              :value="key === chosenSize"
+              :key="key"
+              @click="chooseSize(key)"
```
aabounegm: Event name is "input". That's why the chosen size doesn't get updated
```suggestion
              @input="chooseSize(key)"
```
After that there is the edge case of choosing a size and then switching to a color where this size isn't available; the `chosenSize` remains the same. Need to reset `chosenSize` when choosing a different color

containers/shop/product-modal.vue
VanishMax: It would be fixed after some markup changes. About click-away: we will do it for modals, dropdowns and dialogs after some time. Right now 'X' button is all right

containers/shop/balance.vue
VanishMax: Done

containers/shop/product-card.vue
VanishMax: Done

containers/shop/product-card.vue
VanishMax: Done

store/product.js
VanishMax: Good point. Done

PR #32, opened 2019-08-01
Profile page

VanishMax:
Is made to be changed in a future. Needs a lot of data to work with. For example, timeline container is made properly because of the data unknowledgement.


components/ui/button.vue
```diff
@@ -24,6 +24,10 @@
     </template>
 
     {{ label }}
+
+    <template v-if="chevron">
```
aabounegm: Why would a button normally have a chevron?
Either change the content of the button to be a slot (as would be the case if #30 gets merged), or add another slot just for the icon (named append-icon) so that it can be as generic as possible.

components/profile/radio-group.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <div role="group" class="switches with-labels">
+    <label
+      v-for="item in items"
+      :key="item.id"
+      class="clickable"
+    >
+      <span class="label">{{ item.label }}</span>
+      <div class="radio">
```
aabounegm: Can't the RadioGroup UI component be reused here somehow?
Even if it will require a bit of refactoring, it should be done because the point of UI components are to be as reusable as possible

constants/profile/user.js
```diff
@@ -0,0 +1,7 @@
+export default {
```
aabounegm: This should be in the store as it can change through user interactions and needs to be shared among multiple components

containers/profile/info.vue
```diff
@@ -0,0 +1,68 @@
...
+          outline
+          img="/images/profile/send.svg"
+          label="add Telegram username"
+        />
+
+        <Dropdown :chevron="false" right>
+          <template v-slot:opener>
+            <img src="/images/profile/help-circle.svg" class="icon" />
+          </template>
+          <template>
```
aabounegm: No need to wrap the default slot in `<template>` tags

containers/profile/statistics.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <section class="statistics padded">
+    <header>
+      <h2 class="btn-align">Statistics</h2>
+      <Dropdown class="btn-shift">
```
aabounegm: Will need to be changed when #30 is merged

containers/profile/timeline.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <Timeline :class="{'collapsed': collapsed}" load class="padded">
```
aabounegm: The ES6 object shorthand looks a lot better in cases like this
```suggestion
  <Timeline :class="{collapsed}" load class="padded">
```

components/ui/button.vue
VanishMax: Well, this is not about the icon itself. Chevron has its own class with a transition applied, so each click on the button would rotate the chevron. It is useful

constants/profile/user.js
VanishMax: True. Changing

containers/profile/info.vue
VanishMax: Done

containers/profile/statistics.vue
VanishMax: Done

containers/profile/timeline.vue
VanishMax: Hell, that's great!

components/profile/radio-group.vue
VanishMax: The logic of this radio group is not yet fully done. It'll be much bigger and more difficult to make it reusable. Don't like to too, but have to live with that

PR #33, opened 2019-08-05
Feature/product preview

aabounegm:
Completes task https://trello.com/c/Tjsryc2V/54-product-edit-preview-card
Syncs all data from the varieties being created in a new product with the preview component using Vuex.


store/newProduct.js
```diff
@@ -0,0 +1,81 @@
...
+      state.type = '';
+      state.description = '';
+      state.price = 0;
+      state.inSizes = initialSizes;
+      if(initialSizes) state.quantity = 0;
+      state.varieties.forEach(v => v.images.splice(0, v.images.length));  // to clear the old references
+      state.varieties = [newVariety()];
+      state.varietiesCount = 1;
+    },
+    setName(state, payload) {
```
VanishMax: create a function that accepts name of the field and gives it a payload value

```suggestion
    setField(state, {name, value}) {
      state[name] = value;
    }
```

containers/product-edit/preview.vue
```diff
@@ -18,18 +50,53 @@
       </div>
     </Card>
     <div class="actions">
-      <Button danger label="clear fields" />
+      <Button @click="$store.commit('newProduct/reset')" danger label="clear fields" />
```
VanishMax: Better use mapActions and call the function reset instead of accessing $store

aabounegm: I would've done that if there was more than one action (it's called mapAction**s** and not Action 😜). Not really worth the complication for just one method, imho

PR #48, opened 2019-10-14
Feature/radio buttons

VanishMax:


src/routes/ui.svelte
```diff
@@ -9,6 +9,25 @@
...
+      color: '#FF0000',
+    },
+    {
+      id: 2,
+      label: 'two label',
+      color: '#000000',
+    },
+    {
+      id: 3,
+      label: 'fuck label',
```
illright: Language, please. We are not developing the Linux kernel.

src/containers/profile/notifications.svelte
```diff
@@ -1 +1,121 @@
...
+    projs: {},
+  };
+
+  const changeAll = (e) => {
+    const item = e.detail;
+
+    selected.all = item;
+    selected.store = item;
+    selected.volunteering = item;
+    selected.projs = item;
```
aabounegm: Would be better to use a `for .. in` loop, so that it wouldn't have to be manually changed in case more items are added in the future

src/containers/profile/notifications.svelte
```diff
@@ -1 +1,121 @@
...
+        <p>Push notifications require permission before they can be used.</p>
+      </div>
+      <div class="actions">
+        <Button>prompt for permissions</Button>
+      </div>
+    </div>
+  </header>
+
+  <ul>
+    <li class="section">
```
aabounegm: A lot of code repetition for the `<li>`s. Could be wrapped in a `for` with the unique data in the javascript.

src/containers/profile/notifications.svelte
```diff
@@ -1 +1,121 @@
...
+    const item = e.detail;
+
+    selected.all = item;
+    selected.store = item;
+    selected.volunteering = item;
+    selected.projs = item;
+  };
+
+  const changeOne = (name, e) => {
+    selected[name] = e.detail;
```
aabounegm: Not sure what the UX would say here, but I guess manually choosing one item should clear the "All" option.

src/routes/ui.svelte
```diff
@@ -154,76 +173,40 @@
     </div>
   </Card>
 
-  <!--todo: radios-->
   <Card id="radios">
```
aabounegm: Page is too big. Why not dividing on containers?

src/components/ui/radio-group.svelte
```diff
@@ -0,0 +1,89 @@
...
+
+  let dispatch = createEventDispatcher();
+  const changeRadio = item => {
+    selectedId = item.id + 1;
+    dispatch('change', item);
+  };
+
+  const isWhite = item => {
+    if (!item.color || !isColor) return false;
+    if (item.color.toUpperCase() === '#FFF' || item.color.toUpperCase() === '#FFFFFF') return true;
```
aabounegm: If condition didn't match, will return undefined because of no default/fallback `return false;`
```suggestion
    return item.color.toUpperCase() === '#FFF' || item.color.toUpperCase() === '#FFFFFF';
```

src/components/ui/card.svelte
```diff
@@ -1,16 +1,24 @@
...
-  <slot name="image">
-    {#if img}
-      <img src={img} style={'background: ' + color} class="image" alt="" />
-    {/if}
-  </slot>
   {#if img}
+    {#if imgWrap}
+      <div class="image">
+        <slot name="image">
+          <img src={img} style="{'background: ' + color}" class="image" alt="" />
```
aabounegm: Image size is not bounded by container
image
```suggestion
          <img src={img} style="{'max-width: 100%; background: ' + color}" class="image" alt="" />
```

static/css/global/special.css
```diff
@@ -0,0 +1,11 @@
+.radio .icon.white {
```
aabounegm: Why is this not in the radio-group component?

src/components/ui/card.svelte
VanishMax: Guess it's because different problem. But with your comment, I found it and fixed

src/components/ui/radio-group.svelte
VanishMax: Good notice

src/containers/profile/notifications.svelte
VanishMax: Agree with this good practice. Just was lazy to do it before and afraid if radio-group is not that good for it. But now changed it and it suits perfectly for `for` wrapper

src/routes/ui.svelte
VanishMax: In some near future

src/containers/profile/notifications.svelte
VanishMax: Ah, just understood what you meant. Changed

PR #51, opened 2019-11-12
Filters

VanishMax:
A lot of new updates:
* Checkbox group
* Radio group
* Chip
* Chip group
* Date picker
* UI-conventions
* Filters in project browser and store pages


src/components/ui/text-field.svelte
```diff
@@ -38,29 +38,44 @@
...
-      {min}
-      {max}
-      {maxLength}
-      on:focus={() => dispatch('focus')}
-      on:blur={() => dispatch('blur')}
-      on:input={() => dispatch('input')}
-      on:keyup.enter={() => dispatch('blur')}
-      on:change={(e) => dispatch('change', e.target.value)}
-    />
+    {#if isNum}
```
aabounegm: Why the repetition? Can't type just depend on "isNum" and the other props have default value `null`?

(https://stackoverflow.com/questions/57392773/error-type-attribute-cannot-be-dynamic-if-input-uses-two-way-binding)

src/constants/projects/competences.js
```diff
@@ -1,26 +1,26 @@
 export default [
   {
-    name: 'creativity',
-    checked: true,
+    id: 0,
```
aabounegm: This is a JS file, not JSON. could've simply used 
```js
['creativity', 'critical thinking', '...'].map((label, id) => ({label, id}));
```
😁

src/containers/projects/filters.svelte
```diff
@@ -0,0 +1,56 @@
...
+  import Spots from '@/components/projects/spots.svelte';
+  import Competences from '@/components/projects/competences.svelte';
+  import DateRange from '@/components/projects/date.svelte';
+
+  import orders from '@/constants/projects/order';
+  import {selectedOrder, selectOrder, filters, changeSearch} from '@/store/projects';
+
+  let open = null;
+  const changeOpen = (name) => {
+    if (open !== name || !open) open = name;
```
aabounegm: if `open` was null or undefined, it will not be equal to `name`. Therefore, the `|| !open` is not needed imho.

src/store/store.js
```diff
@@ -0,0 +1,47 @@
...
+export function changeColors(value) {
+  changeFilter('colors', value);
+}
+
+export function changeIsColorless() {
+  let {isColorless} = get(filters);
+  changeFilter('isColorless', !isColorless);
+}
+
+export function changePrice(value, isHigh = false) {
```
aabounegm: How about just pass an object with both parameters instead of a parameter that defines how to interpret the other parameter? 😬 
```suggestion
export function changePrice({low=0, high}) {
```

src/store/projects.js
```diff
@@ -0,0 +1,38 @@
...
+function changeFilter(name, value) {
+  filters.update((state) => ({...state, [name]: value}));
+}
+
+export function changeSearch(value) {
+  changeFilter('search', value);
+}
+
+export function changeSpots(value) {
+  if (value > 0) changeFilter('spots', value);
```
aabounegm: A simpler way would be to write 
```suggestion
  if (value > 0) value = 1;
```
Or even 
```suggestion
  value = Math.max(1, value);
```
and just use the variable `value` right away

src/components/ui/text-field.svelte
VanishMax: Right. Managed doing sp

src/store/projects.js
VanishMax: Oh maaan, you're a math profi

aabounegm: I just don't like to repeat myself as much as possible 😁

src/components/ui/checkbox-group.svelte
```diff
@@ -0,0 +1,78 @@
...
+</style>
+
+<div role="group" class:with-labels={labeled} class={classname}>
+  {#each items as item (item.id)}
+    <label class:clickable={labeled} class={labelclass}>
+      <div class:colored={colored} class:round={colored} class="checkbox {checkboxclass}">
+        <input
+            on:change={() => changeCheckbox(item)}
+            type="checkbox"
+            checked={selected.find(x => x.id === item.id)}
```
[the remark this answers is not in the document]
VanishMax: It's a perfect one

src/constants/projects/competences.js
VanishMax: This comment has just changed my mind. Will add a task for this to use it in all -group components later

src/store/store.js
VanishMax: divided to two functions

PR #52, opened 2019-11-27
Refactor UI component system

VanishMax:
Make sure all component follow industrial-oriented UI-conventions, add plenty of logic to checkbox, radio and chip group, add features such as click-away and ripple, apply it to the whole project


src/components/store/card.svelte
```diff
@@ -12,6 +12,7 @@
   export let short = false;
 
   $: selected = varieties[0];
+  // $: console.log(selected);
```
aabounegm: Delete the line instead of commenting it out

src/components/ui/utils/parse-items-array.js
```diff
@@ -0,0 +1,16 @@
+export default (arr, uniqueKey) => {
+  if (Array.isArray(arr) && arr[0]) {
+    let first = arr[0];
+    if (first && typeof first === 'object') {
```
aabounegm: Redundant check for `first`. Already check in the `arr[0]` part.

src/components/ui/utils/parse-items-array.js
```diff
@@ -0,0 +1,16 @@
+export default (arr, uniqueKey) => {
+  if (Array.isArray(arr) && arr[0]) {
```
aabounegm: A lot of nesting `if`s. Use guard clauses instead.
Example:
```js
if(!Array.isArray(arr))
    return arr;
```

src/components/ui/switch.svelte
```diff
@@ -1,34 +1,51 @@
...
 
-  {#if twoState}<span class="state first">{first}</span>{/if}
-  <div class="{(twoState ? 'two-state' : 'ml-2') + ' switch'}">
-    <div class="knob" />
+  {#if isTwoState}
+    <slot name="first-state">
+      <span class={firstclass}>{first}</span>
+    </slot>
+  {/if}
+  <!--TODO: add scoped styles for .two-state and .ml-2 or a way to rewrite these classes -->
```
aabounegm: Is this supposed to be done in this PR or later?

src/containers/ui/chips.svelte
```diff
@@ -0,0 +1,53 @@
...
+  <hr />
+
+  <div class="showcase">
+    <div role="group" style="display: flex; flex-wrap: wrap;">
+      <Chip name="chip-2" small>XS</Chip>
+      <Chip name="chip-2" small value={true}>S</Chip>
+      <Chip name="chip-2" small disabled>M</Chip>
+      <Chip name="chip-2" small outline>L</Chip>
+      <Chip name="chip-2" small outline>XL</Chip>
+    </div>
```
aabounegm: Isn't the `ChipGroup` there exactly for this purpose?

src/components/ui/dropdown.svelte
```diff
@@ -1,31 +1,55 @@
...
     {/if}
   </Button>
-  <div class:right-edge={right} class="dropdown {dropdownclass}">
+  <div class:right-edge={isRight} class={dropdownclass} bind:this={dropNode}>
     {#if nowrap}
       <slot />
     {:else}
-      <div class="relative-wrapper {wrapperclass}">
-        <Button on:click={toggle} normal round classname="close">
+      <div class={wrapperclass}>
```
aabounegm: Wrapping should add padding to content because it gets clipped at the corners. The close button should not affect the content.
image

VanishMax: Css-oriented problem is for the markup repo

src/components/ui/button.svelte
```diff
@@ -34,6 +27,18 @@
...
+  const getRippleColor = () => {
+    if (ripple) {
+      if (isFilled) {
+        return 'rgba(255, 255, 255, .35)';
+      } else {
+        if (isDanger) return 'rgba(186, 3, 3, .25)';
+        else return 'rgba(56, 120, 0, .25)';
+      }
+    }
+    return null;
```
aabounegm: Same comment here: using guard clauses can make it only one level of nesting
```suggestion
    if(!ripple)
      return null;
    if(isFilled)
      return 'rgba(255, 255, 255, 0.35)';
    if(isDanger)
      return 'rgba(186, 3, 3, 0.25)';
    return 'rgba(56, 120, 0, 0.25)';
```

aabounegm: Also, the function is used in a line before it is declared (using `const`!).
And it can be inlined using a reactive block instead.
```js
$: {
  if(!ripple)
    rippleColor = null;
  // ...
}
```

PR #53, opened 2019-11-29
New product page

VanishMax:


src/store/item.js
```diff
@@ -1,4 +1,13 @@
 import { writable } from 'svelte/store';
+import colorOptions from '@/constants/item/colors';
+
+export const colors = writable(colorOptions);
+export const customColors = writable({});
+export function addNewColor(id, color) {
+  customColors.update(x => {
```
aabounegm: Check if this color already exists first before adding it

src/store/item.js
```diff
@@ -1,4 +1,13 @@
...
+    return {...x, [id]: { color }};
```
aabounegm: This object looks exactly like an array. 
image
The structure should be changed to an array. The function is called "add", which means it should append to the end. Currently, with the index being passed, you can only add 1 extra color for each variety

src/store/item.js
```diff
@@ -1,4 +1,13 @@
 import { writable } from 'svelte/store';
+import colorOptions from '@/constants/item/colors';
+
+export const colors = writable(colorOptions);
+export const customColors = writable({});
```
aabounegm: Why do we need 2 separate structures?

src/utils/read-files.js
```diff
@@ -0,0 +1,18 @@
+// Get array of files ([File, File]), open each asynchronously
+// and return array of strings that should be passed as src:
+// <img src={images}>
+
+export default async (arr) => {
+  let urls = [];
+  for (let i = 0; i < arr.length; i++) urls.push(await readFileAsync(arr[i]));
```
aabounegm: This way, it waits until the current file is completely processed to move on to the next. Better use `Promise.all` with `map`
```suggestion
  urls = await Promise.all(arr.map(readFileAsync));
```
Basically, the whole function can be written in one line:
```js
export default (arr) => Promise.all(arr.map(readFileAsync));
```

src/constants/item/colors.js
```diff
@@ -0,0 +1 @@
+export default ['#FF0000', '#00FF00', '#0000FF', '#000000'].map(x => ({color: x}));
```
aabounegm: Why does it need to be mapped to objects?

src/components/item/form-variety.svelte
```diff
@@ -1,44 +1,61 @@
...
-  const sizes = ['XS', 'S', 'M', 'L' , 'XL', 'XXL'];
-  const variety = $item.varieties[index];
+  let colorPicker = null;
+
+  $: variety = $item.varieties[index];
+  const chooseColor = e => {
+    changeVarietyField(index, 'color', e.detail.color);
+  };
+
+  const openColorPicker = () => colorPicker.click();
```
aabounegm:
```suggestion
  const openColorPicker = colorPicker.click;
```
(which doesn't need an extra function, imo)

src/components/ui/dropzone.svelte
```diff
@@ -0,0 +1,71 @@
...
+  export let disabledQuerySelector = null;
+
+  $: files = value;
+
+  let filePlatform = null;
+  const dispatch = createEventDispatcher();
+  const fileClick = (e) => {
+    if (disabled) return;
+    if (disabledElements.length && disabledElements.some(el => {
+      return el.contains(e.target);
```
aabounegm: Useless body. `param => { return f(param); }` is equivalent to `param => f(param)`.
Just write `.some(el => el.contains(e.target))`

src/components/ui/dropzone.svelte
```diff
@@ -0,0 +1,71 @@
...
+  const upload = newFiles => {
+    for (let i = 0; i < newFiles.length; i++) {
+      const file = newFiles[i];
+      if (maxSize) {
+        if (file.size > maxSize * 1024 * 1024) {
+          dispatch('error', 'FILE_SIZE_EXCEEDED');
+          return;
+        }
+      }
+      multiple ? files.push(file) : files = [file];
```
aabounegm: Confusing expression. Use normal if statement
```suggestion
      if(multiple)
        files.push(file)
      else
        files = [file];
```

src/routes/_error.svelte
```diff
@@ -1,18 +1,48 @@
...
+    <div class="text">
+      <h1>Nothing here</h1>
+      <p class="subtitle">That rabbit must have given you the wrong lead.</p>
+      <Button href="/" isFilled>
+        <svg src="/images/icons/home.svg" class="icon mr" />
+        take me home
+      </Button>
+    </div>
+    <img src="/images/404/web-traveller.svg" width="240" height="387" class="picture" alt="">
+    <h2>Feeling lucky?</h2>
```
aabounegm: This heading is not visible on desktop.
The links' images do not represent the destination, and there is no additional label/description

src/components/item/drag-dropzone.svelte
```diff
@@ -0,0 +1,69 @@
...
+  export let index;
+  export let files;
+
+  let images = [];
+  $: (async() => images = await openFiles(files))();
+
+  const changeFiles = async e => changeVarietyField(index, 'images', e.detail);
+  const removeImage = (data) => {
+    let pos = images.indexOf(data);
+    if (pos > -1) {
```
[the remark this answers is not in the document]
VanishMax: In this case, guard clause is not that well-looking and efficient, but changed

src/components/item/drag-dropzone.svelte
```diff
@@ -0,0 +1,69 @@
...
+      changeVarietyField(index, 'images', files.filter((x, i) => i !== pos));
```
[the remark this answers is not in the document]
VanishMax: Nah, we don't need to modify `files` array in here. And efficiency is arguable though we have to find index with `indexOf` first that searches almost like `if` in filter.

src/store/item.js
VanishMax: Right now it doesn't matter, because we don't have real color-picker and all the colors are not well-defined. The implemented logic is a "close interpretation of real one"

src/store/item.js
VanishMax: Sure about the object??
image
Anyway, it is passed to one variety just because it is logical. If we try to just extend the `colors` array, we'll get aloooooooot of color in there due to `input type=color` work and no ability to delete them from ui.

src/store/item.js
VanishMax: They will be merged after product creation. As I wrote in the previous comment, it distinguishes between pre-defined colors and colors for varieties without mixing indexes.

src/utils/read-files.js
VanishMax: Oh, thanks. That's really good

src/constants/item/colors.js
VanishMax: The `checkbox-group` searches for color field in an array of objects. It supports array of strings only for labels, but not colors.

src/components/item/form-variety.svelte
VanishMax: You have just wrote the assignment of `null` variable `colorPicker`'s undefined property `click`. Constant function is needed for imitating the click on the input to open the color picker.

src/routes/_error.svelte
VanishMax: Markup problems 🤷‍♂️

src/store/item.js
aabounegm: I added a `console.log` right before that return with the same expression, and that's what I got.

src/components/item/form-variety.svelte
aabounegm: Ah, I didn't notice this. Sorry

PR #57, opened 2019-12-27
Create project

VanishMax:
Page for project creation consists of 4 steps.
What's not yet implemented
* Pre-defined project templates (student party or olympiad)
* Saving images to backend (localStorage cannot keep files)
* Autocomplete with real data
* Creation and finding (and jumping to) incomplete fields


illright:
Here's a checklist of things to fix:
- [ ] Fix the "start over" button (also applies to pressing the back button)
- [ ] Fix the label in the dropzone to say 16:9 instead of 1:1
- [ ] Title-case the page `<title>`: Create Project
- [ ] For hourly activities, multiply 70 by the working hours and show that as the "reward" instead of "70 ipts/h" (I'm talking about a saved card that can be dragged)
- [ ] Disallow negative values for the people picker
- [ ] Replace "1 people" for "1 person"
- [ ] Add the "discard changes" button to the cards being edited (see [/view-project-creator.html](https://illright.github.io/innopoints-markup/view-project-creator.html))
- [ ] When adding a new feedback question, the text of the previous question remains in the form
- [ ] Deleting a feedback question causes the popup to close
- [ ] The calendars don't close automatically when the choice is made
- [ ] The application calendar shouldn't have "start date" and "end date" toggles, since it's a single date
- [ ] When some of the required fields in the activity form are not filled in, the save button doesn't save, and the website doesn't tell what needs to be filled out
- [ ] For the hourly reward, do not allow selecting the reward rate (in that case it is fixed to 70)
- [ ] When the date (or date selection) is made, the selector button text doesn't update to the selected date
- [ ] Disable the people field when the "the more, the better" checkbox is checked
- [ ] The competence selector allows picking more than 3 competences
- [ ] The "some required fields are still empty" doesn't disappear when all the fields are filled out (note: for steps 1 and 2 it must disappear when all the fields **for that step** are filled out, for step 3 it must disappear only if all the fields for the project are filled out
- [ ] The textfield on step 3 doesn't have a label (doesn't match the markup)
- [ ] The "where" button next to the "some required fields are still empty" field isn't there (when it's there, it should be a link to the step which has such fields, the first one, if several exist)
- The "create project" button must be disabled if there are empty required fields
- The "create project" button doesn't do anything :)

illright:
The markup has style changes of the UI kit, adding the error states for form fields. To signify which required fields were not filled in, add the `.error` class to the `.form-field`.

src/containers/projects/new/step-0.svelte
```diff
@@ -0,0 +1,95 @@
+<script>
+  import { fade } from 'svelte/transition';
+  import { onMount } from 'svelte';
+
+  import Button from 'ui/button.svelte';
+  import DraftCard from '@/components/projects/new/draft-card.svelte';
+  import {drafts, getDrafts, createProject} from '@/store/new-project';
+  onMount(() => getDrafts());
```
aabounegm:
```suggestion
  onMount(getDrafts);
```

src/containers/projects/new/step-1.svelte
```diff
@@ -0,0 +1,86 @@
...
+  import Button from 'ui/button.svelte';
+  import Dropzone from 'ui/dropzone.svelte';
+  import {project, changeField, save} from '@/store/new-project';
+  import {readFileAsync} from '@/utils/read-files';
+
+  $: file = $project.image;
+  let image = null;
+  $: (async() => image = file ? await readFileAsync(file) : null)();
+
+  const addImage = async e => changeImageAndSave(e.detail[0]);
```
aabounegm: why `async`?

src/containers/projects/tagline.svelte
```diff
@@ -9,5 +14,13 @@
       If you’re seeking an opportunity to contribute to the wonderful world of
       Innopolis projects, you’re in the right place!
     </div>
+    {#if $isAuthed && $user.isAdmin}
```
aabounegm: Anyone, not only admins, can create a project

src/layouts/auth.svelte
```diff
@@ -2,18 +2,28 @@
...
   import Footer from '@/components/footer.svelte';
   import Button from 'ui/button.svelte';
-  import { isAuthed, logIn } from '@/store/user';
+  import { isAuthed, logIn, user } from '@/store/user';
 
+  export let classname = '';
   export let title;
+  export let adminsOnly = false;
+
+  $: condition = $isAuthed && (!adminsOnly || (adminsOnly && $user.isAdmin));
```
aabounegm:
```suggestion
  $: condition = $isAuthed && (!adminsOnly || $user.isAdmin);
```

src/routes/projects/new.svelte
```diff
@@ -0,0 +1,39 @@
...
+<Layout classname="step{$step}" title="Create project">
+  {#if $step === 0}
+    <StepZero />
+  {:else if $step === 1}
+    <StepOne />
+  {:else if $step === 2}
+    <StepTwo />
+  {:else if $step === 3}
+    <StepThree />
+  {/if}
```
aabounegm: How about `<svelte:component this={steps[$step]}/>`, where `steps = [StepZero, StepOne, ...]`?

src/components/projects/new/autocomplete.svelte
```diff
@@ -0,0 +1,44 @@
+<script>
+  import Dropdown from 'ui/dropdown.svelte';
+  import Button from 'ui/button.svelte';
+
+  export let classname = 'autocomplete-field';
+  export let value = [];
```
aabounegm: It's an array. better call it "values"

src/components/projects/new/autocomplete.svelte
```diff
@@ -0,0 +1,44 @@
...
+  <Dropdown bind:value={open} classname="dropdown-shell open" nowrap>
+    <input
+      slot="handle"
+      bind:value={search}
+      on:focus={() => focus = true}
+      on:blur={() => focus = false}
+      on:input={() => open = true}
+    />
+    {#each options.slice(0, 5) as option (option)}
+      {#if value.indexOf(option) === -1}
```
aabounegm:
```suggestion
      {#if !value.includes(option)}
```

src/components/projects/new/draft-card.svelte
```diff
@@ -0,0 +1,31 @@
...
+
+  const choose = e => {
+    e.preventDefault();
+    chooseProject(date);
+  };
+</script>
+
+<div class="card" in:slide="{{duration: 300}}">
+  <div class="title">{title || 'Untitled'}</div>
+  {#if new Date(date) instanceof Date && !isNaN(new Date(date))}
```
aabounegm: How can an object instantiated using `new Date()` _not_ be an `instanceof Date`?

src/containers/projects/new/step-1.svelte
VanishMax: Because we have to resolve the promise

src/components/projects/new/autocomplete.svelte
VanishMax: UI-conventions disagree

src/components/projects/new/draft-card.svelte
VanishMax: Simply: https://stackoverflow.com/a/1353711/7442294

src/containers/projects/new/step-1.svelte
aabounegm: Which promise? `changeImageAndSave` is a synchronous function, and you're not using any `await`s

src/components/ui/checkbox-group.svelte
```diff
@@ -12,14 +12,18 @@
...
 
   let dispatch = createEventDispatcher();
-  const changeCheckbox = item => {
+  const changeCheckbox = (item, e) => {
     if (selected.some(x => x.id === item.id)) selected = selected.filter(x => x.id !== item.id);
-    else selected.push(item);
+    else {
+      if (!max || selected.length < max) selected.push(item);
+      else e.target.checked = false;
+    }
```
aabounegm: I think it would be better in terms of UX to actually disable the unchecked checkboxes themselves when 3 are checked.
Something along the lines of `disabled={!selected.some(x => x.id === item.id) && max != null && selected.length >= max}` on the input element (and maybe refactor the first clause into a function and the other two into a reactive variable? that line is a bit long)

src/containers/projects/new/step-3.svelte
```diff
@@ -1,9 +1,12 @@
 <script>
   import { fade } from 'svelte/transition';
+  import { onMount } from 'svelte';
   import StepHeader from '@/components/projects/new/step-header.svelte';
   import BottomNavigation from '@/components/projects/new/bottom-navigation.svelte';
   import Autocomplete from '@/components/projects/new/autocomplete.svelte';
+  import {checkProject} from '@/store/new-project';
 
+  onMount(() => checkProject());
```
aabounegm: Again, `onMount(checkProject);`. You already changed the other one)

src/store/new-project.js
```diff
@@ -88,14 +96,45 @@ export const changeActivityField = (index, field, value) => {
...
+export const findErrors = () => {
+  checkProject();
+  const errs = get(errors);
+  if (errs.includes('name') || errs.includes('organizer')) goto('/projects/new?step=1');
+  else if (errs.includes('activities')) goto('/projects/new?step=2');
+};
+
+export const endProject = () => {
+  checkProject();
+  if (!get(errors).length) {
```
aabounegm: Same about guard clauses and `length`

src/containers/projects/new/step-1.svelte
VanishMax: Oh, sorry, I was looking at the readFileAsync function above. Right

src/components/ui/checkbox-group.svelte
VanishMax: Yeah, maybe. But not now)

src/components/projects/new/new-activity.svelte
```diff
@@ -51,6 +55,14 @@
     changeAndSave(field, val);
     if ((two && val.start && val.end) || (!two && val)) dropdowns[field] = false;
   };
+
+  const create = (isEdit = false) => {
+    errors = checkActivity(activity);
+    if (!errors.length) {
```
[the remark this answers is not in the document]
VanishMax: But these kind of conditional nesting make full sense of the conditional flow. So, it's just the matter of preferences. I don't mind you writing guard clauses)

src/store/new-project.js
VanishMax: Here it helps, agree

PR #65, opened 2020-01-12
Revive store

illright:
The /store page is back to operation.

What's still not there:
- product cards (those will be fixed in a later PR)
- session information like balance (this will be added in an even later PR)


src/routes/store/index.svelte
```diff
@@ -1,18 +1,89 @@
...
+
+  let balance;
+  let order = orderOptions[0];
+  let orderLabel = orderLabels[0];
+  let currentPage = 1;
+
+  let filterElement;
+
+  try {
+    balance = $session.user.balance;
```
VanishMax: I bet you don't ever have `session.user` because it should be sent by sapper middleware from *src/server.js* just like it is explained here. 
Then imagine you have installed **express-session** and seeded it to sapper. The problem is how to save the session in the first place cause `user` object will still be empty. One solutions is to save it when you have authenticated via client (in `onMount` function), and then use it for future usage without forgetting to delete session after log out.

src/routes/store/index.svelte
```diff
@@ -1,18 +1,89 @@
...
+    { orderBy: 'addition_time', order: 'desc' },
+    { orderBy: 'price', order: 'asc' },
+    { orderBy: 'price', order: 'desc' },
+  ];
+  const orderLabels = [
+    'newest first',
+    'newest last',
+    'cheapest first',
+    'cheapest last',
+  ];
```
VanishMax:
```
import orderOptions from '@/constants/store/orderOptions.js'
import orderLabels from '@/constants/store/orderLabels.js'
```
?

src/routes/store/index.svelte
```diff
@@ -1,18 +1,89 @@
 <script context="module">
+  import * as api from '@/utils/api.js';
+  const productLimit = 24;
+
   export async function preload() {
-    const res = await this.fetch('/store.json');
-    return {items: await res.json()};
+    const { pages, data } = await api.get(`/products?limit=${productLimit}`);
```
VanishMax: Here is the point when it is really better to use `this.fetch` Sapper's wrapper instead of fetch API or node-fetch ([explained](https://sapper.svelte.dev/docs#this_fetch)). The main comment is in _@/utils/api.js_

src/routes/store/index.svelte
```diff
@@ -1,18 +1,89 @@
...
+  } catch (e) {
+    balance = 1;
+  }
 
-  export let items;
+  function updateProducts(filtering) {
+    let query = [];
+    if (filtering.searchQuery) {
+      query.push(`q=${filtering.searchQuery}`);
+    }
```
VanishMax: There is a _'@/utils/generateQuery.js'_ function that generates a URL-query in a quick and safer for XSS-attacks way. I don't mind using exactly that function, but some of the lines will show how to do it with prettier code.

src/routes/store/index.svelte
illright: I don't even know why you review this) I've said in the description that I didn't do session interaction yet, this is merely a placeholder

src/routes/store/index.svelte
illright: fair enough, will replace

src/utils/api.js
```diff
@@ -0,0 +1,67 @@
...
+  return fetch(url, options)
+		.then(r => r.text())
+		.then(json => {
+			try {
+				return JSON.parse(json);
+			} catch (err) {
+				return json;
+			}
+		});
+}
```
aabounegm: mix of spaces and tabs again

src/utils/api.js
```diff
@@ -0,0 +1,67 @@
...
+
+  return fetch(url, options)
+		.then(r => r.text())
+		.then(json => {
+			try {
+				return JSON.parse(json);
+			} catch (err) {
+				return json;
+			}
+		});
```
aabounegm: Why not simply `r => r.json()`? I don't think there is any endpoint in the backend that doesn't return JSON (or no payload). And even if, better check the response headers for Content-Type

src/utils/api.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+
+export function get(path) {
+  if (API_HOST == null) {
+    throw new Error('The API host is undefined.');
+  }
```
aabounegm: Maybe this can be checked once right after declaring `API_HOST`? It's an environment variable; shouldn't be reactive

src/routes/store/index.svelte
```diff
@@ -1,18 +1,89 @@
 <script context="module">
+  import * as api from '@/utils/api.js';
+  const productLimit = 24;
+
   export async function preload() {
-    const res = await this.fetch('/store.json');
-    return {items: await res.json()};
+    const { pages, data } = await api.get(`/products?limit=${productLimit}`);
+    const colors = await api.get('/colors');
```
aabounegm: Same as comment in ./index.json.js

src/routes/store/index.svelte
```diff
@@ -1,18 +1,89 @@
...
+    if (filtering.minPrice != null) {
+      query.push(`min=${filtering.minPrice}`);
+    }
+    if (filtering.maxPrice != null) {
+      query.push(`max=${filtering.maxPrice}`);
+    }
+    let excludedColors = filtering.excludedColors.map(color => color.slice(1));
+    if (filtering.colorlessExcluded) {
+      excludedColors.push('null');
+    }
```
aabounegm: There is the utils/generateQuery module precisely to format the parameters in a query. no need to do it here manually

src/routes/store/index.svelte
```diff
@@ -1,18 +1,89 @@
...
+      query.push(`excluded_colors=${excludedColors.join(',')}`);
+    }
+    query.push(`order_by=${filtering.order.orderBy}`);
+    query.push(`order=${filtering.order.order}`);
+    let queryString = `limit=${productLimit}`;
+    if (query) {
+      queryString += '&' + query.join('&');
+    }
+    queryString += `&page=${currentPage}`;
+    console.log(queryString);
```
aabounegm: no `console.log`s in production code

src/containers/store/filters.svelte
```diff
@@ -1,107 +1,181 @@
...
+                   || (filtering.maxPrice != null)
+                   || (filtering.excludedColors.length !== 0)
+                   || (filtering.colorlessExcluded));
+
+  colors.forEach(colorObject => { colorObject.checked = true; });
+  let colorlessValues = [{ value: null, checked: !filtering.colorlessExcluded }];
+
+  const dispatch = createEventDispatcher();
+
+  let searchThrottle = null;
```
aabounegm: This is actually debounce and not throttle

src/containers/store/filters.svelte
```diff
@@ -1,107 +1,181 @@
...
-                  isNoSpinner
-                  value={$filters.priceLow}
-                  on:input={(e) => changeLowPrice(e.detail)}
-                  on:change={(e) => changeLowPrice(e.detail)}
+                type="number"
+                isNoSpinner
+                value={filtering.minPrice}
+                max={filtering.maxPrice}
+                on:input={changeMinPrice}
               />
```
aabounegm: Also add `min="0"`

src/components/common/pagination.svelte
```diff
@@ -0,0 +1,168 @@
...
+
+  function generateButtonList() {
+    let list = [];
+    if (currentPage !== 1) {
+      list.push(1);
+    }
+
+    if (currentPage - 1 == 3) {
+      list.push(currentPage - 2);
+    } else if (currentPage - 1 > 3) {
```
aabounegm: Why all the weird arithmetics? These can simply be
`currentPage == 4`
`list.push(2)`
`currentPage > 4`

src/components/common/pagination.svelte
```diff
@@ -0,0 +1,168 @@
...
+          autofocus
+          on:change={(evt) => goTo(+evt.detail)}
+        />
+      {:else}
+        <Button classname="page" isNormal on:click={activateRightInput}>...</Button>
+      {/if}
+    {/if}
+    {#if currentPage !== pages}
+      <Button classname="page" isNormal on:click={() => goTo(pages)}>{pages}</Button>
+    {/if}-->
```
aabounegm: Delete commented code

src/components/common/pagination.svelte
illright: Merely for symmetry with the other edge, `pages`, where such simplifications cannot be made.

src/containers/store/filters.svelte
illright: Technically, [this is neither](https://css-tricks.com/the-difference-between-throttling-and-debouncing/). It's just delaying the call. Idk what to best call this technique.

src/routes/store/index.svelte
illright: It's far from being production code) but this is fixed, anyway

src/components/store/balance.svelte
```diff
@@ -0,0 +1,10 @@
+<script>
+  export let value = null;
+</script>
+
+{#if value != null}
+  <div class="balance">
+    You have {value}
+    <svg src="images/innopoint-sharp.svg" alt="innopoint{value !== 1 ? 's' : ''}" />
```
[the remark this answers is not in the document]
illright: It returns the word with the number, which would be duplicated here, as the number is outside the tag

src/components/ui/check-chip-group.svelte
```diff
@@ -25,12 +25,13 @@
   }
 
   const dispatch = createEventDispatcher();
-  function checkLimit(evt) {
+  function checkLimit(evt, value) {
     if (maxChecked !== null && currentChecked + evt.target.checked > maxChecked) {
       evt.preventDefault();
-      return false;
+    } else {
+      value.checked = !value.checked;
```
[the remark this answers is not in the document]
illright: You'd think. But it didn't work without this line, so idfk

src/utils/api.js
illright: I just wanted the server to run even without this variable (for edge debug cases where we aren't actually making requests). Idk if you think this is worth it, let me know

src/utils/api.js
illright: This code was copy-pasted. And it seems alright to me, inspecting the Content-Type in the second callback would be impossible and putting it in the first seems kinda meh.

src/routes/store/index.svelte
illright: I'll be honest here, when I copied the code for `api.js`, I thought it was making use of `this.fetch`. Now that I look at it, I see that it doesn't. Moreover, I thought I was using that `store.json.js` frontend endpoint, but I now see that I'm not. So far it works and I don't know how far this will take us, and if changing any of this is really necessary. That server endpoint looks a little redundant to me. 

Anyway, bottomline: just stick with it, if we encounter issues, we will fix them at the spot.

src/containers/store/filters.svelte
aabounegm: I'm not sure why you say it's neither. Perhaps this demo will make it clearer: http://demo.nimius.net/debounce_throttle/

illright: Oh, yeah, right, sorry. Will rename

src/utils/api.js
aabounegm: Why would you need to run the server without this variable existing, not even empty? The server reads it from the .env file anyways, so you don't have to manually put it there. Also, maybe another variable like `DEBUG=1` for these cases?
Anyways, I think this is a lot of useless code repetition, preventing the function from being much simpler.

src/components/ui/check-chip-group.svelte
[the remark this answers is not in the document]
aabounegm: Maybe it is about the order of calls. This is an event handler, so it makes sense it waits for it to return before changing the bound variable so that if the function calls `preventDefault` or returns false, it doesn't change the variable.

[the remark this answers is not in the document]
illright: Well, this works, I figured I wouldn't put too much thought to it)

src/utils/api.js
aabounegm: What isn't alright is that it will return either JSON or just a string (probably HTML, if something is not properly handled in backend), when it should just return JSON. If it didn't, it should cause an error so that we find it and fix it rather than having something just not working properly and we spend hours trying to find the root cause

src/routes/store/index.svelte
VanishMax: Then why add a useless line? PR should be fully working without any errors. You can make it smaller (f.e. just making auth work or displaying product card using preloading), but not a really wide PR that doesn't work

illright: It's fully working without errors. This if is merely a reminder to get back to. That's for your first two sentences, the last one i didn't really understand

PR #68, opened 2020-01-13
Revive projects (2nd attempt)

illright:
Let's just have another PR and forget all about the first one.
Reason: sorry lmao

Old PR message:
> Get the /projects page back to operation.
> What's still not there:
> - the date picker is unrefactored and not yet added (thus the commented code, bear with it as of now, Abdelrahman)
> - the project cards are most likely not working, will be fixed in a later PR


illright:
@illright @VanishMax awaiting your review

PR #69, opened 2020-01-15
Auth

illright:
Reconstruct the authentication. Allow SSRing the required content directly into the pages on the initial request.

Bonus: work around the notification center


illright:
lol, 69

src/components/common/notification-center.svelte
```diff
@@ -0,0 +1,51 @@
+<script>
+  import { onMount } from 'svelte';
+  import Dropdown from 'ui/dropdown.svelte';
+  import Button from 'ui/button.svelte';
+  import * as api from '@/utils/api.js';
+
+  export let notifications = null;
+  $: unread = notifications != null && notifications.some(notification => !notification.is_read);
+
+  onMount(async () => {
```
VanishMax: Was it intended to load only with the page load? Maybe consider loading on dropdown opening?

src/components/common/notification-center.svelte
```diff
@@ -0,0 +1,51 @@
...
+
+<Dropdown
+  dropdownclass="notification-center"
+  handleclass="round"
+  chevron={false}
+  nowrap
+  let:toggle={toggle}
+>
+  <div slot="label" class="badge" class:hidden={!unread}>
+    <svg src="images/icons/bell.svg" class="icon" />
```
VanishMax: It is really better to load from root like `/images/icons/...`, because in the nested pages it will look for, for example, '/item/3/images/icons/...`

src/components/common/notification-center.svelte
```diff
@@ -0,0 +1,51 @@
...
+  </div>
+  <Button isNormal isRound classname="close" on:click={toggle}>
+    <svg src="images/icons/x.svg" class="icon" />
+  </Button>
+  <div class="title">Notifications</div>
+  {#if notifications == null}
+    <div class="empty small">
+      <div class="icon">
+        <div class="lds-ellipsis"><div></div><div></div><div></div><div></div></div>
```
VanishMax: Wtf?
Loop, may be?

src/components/home/header.svelte
```diff
@@ -1,21 +1,26 @@
 <script>
+  import { stores } from '@sapper/app';
   import Button from 'ui/button.svelte';
-  // import NotificationCenter from '@/components/profile/notification-center.svelte';
-  import { isAuthed, login, user } from '@/store/user';
+  import NotificationCenter from '@/components/common/notification-center.svelte';
+  import { login, loginCheat } from '@/utils/auth.js';
+  import { ENV } from '@/constants/env.js';
+
+  const loginFunction = (ENV === 'development' ? loginCheat : login);
```
VanishMax: Repetition (Here and in main header). Choosing the *loginFunction* could be performed explicitly in *auth.js*

src/components/common/notification-center.svelte
illright: Nah. If there are new notifications, it should show a badge

src/constants/env.js
```diff
@@ -1,3 +1,5 @@
 export const API_HOST = process.env.SAPPER_APP_API_HOST;
 export const AUTH_HOST = process.env.SAPPER_APP_AUTH_HOST;
 export const FRONTEND_BASE = process.env.SAPPER_APP_FRONTEND_BASE;
+export const ENV = process.env.SAPPER_APP_ENVIRONMENT;
```
VanishMax: There is a `process.env.NODE_ENV` variable that is used in the *rollup.config.js*. Is it possible not to create more environment variables?

src/components/common/notification-center.svelte
illright: It is really better not to alias routes all over the place, but as for icons, I prefer to keep them slash-less

src/components/common/notification-center.svelte
illright: This is copy-pasted code from https://loading.io/css/, I really don't see a reason to change it

src/routes/_layout.svelte
```diff
@@ -1,8 +1 @@
-<script>
-  import {onMount} from 'svelte';
-  import {changeUser} from '@/store/user';
-  import request from '@/utils/request';
-  onMount(async () => changeUser(await request(fetch, '/account', 'GET')));
-</script>
-
 <slot />
```
VanishMax: No more need for this file if it is just a slot

src/components/common/notification-center.svelte
```diff
@@ -0,0 +1,51 @@
+<script>
+  import { onMount } from 'svelte';
+  import Dropdown from 'ui/dropdown.svelte';
+  import Button from 'ui/button.svelte';
+  import * as api from '@/utils/api.js';
+
+  export let notifications = null;
+  $: unread = notifications != null && notifications.some(notification => !notification.is_read);
```
aabounegm: `unread` will be false if `notifications == null`. If there are no notifications, then there are no unread notifications
Using `||` will also short-circuit if they are null
```suggestion
  $: unread = notifications == null || notifications.some(notification => !notification.is_read);
```

src/constants/env.js
illright: Idk. I don't trust that variable, so I've created my own, the one that I can easily manipulate

aabounegm: what exactly do you not trust about it? it is 'development' if running through `npm run dev`, and 'production' if running `npm run build`

src/components/common/notification-center.svelte
illright: If notifications are null, then the first condition will fail and `false` will be returned without evaluating the second condition, therefore, there will not be unread notifications. Feels like the same thing

aabounegm: Oh. seems like I had it the other way around. nevermind.

src/constants/env.js
illright: Well, I didn't know that and didn't bother finding out. Besides, we may set it to a heroku-specific value. I like control

src/components/common/notification-center.svelte
aabounegm: Is the way it looks not enough reason?

illright: There isn't even a built-in `range` in JS as in Python. Do you really want me to import a generator from utils just because you don't like to see four plain divs in a row?

PR #70, opened 2020-01-16
Build the content of the notification center

illright:


src/components/ui/button.svelte
```diff
@@ -55,6 +57,7 @@
     {disabled}
     class="btn {classes.join(' ')}"
     on:click on:mousedown on:mouseup
+    title="{tooltip}"
```
VanishMax: Title is not such useful attribute that needs real prop. When I firstly saw it, the idea was born:
You can pass props as {...$$props} but this is not so good, cause in some components (and even in button) it would lead to attribute-spamming: `<button lalka="kek" required nothing="false"` etc. So, it would be cool to filter props somehow. Try to dig into it

src/components/common/notification-center.svelte
```diff
@@ -46,6 +58,39 @@
...
+          <Button
+            isRound
+            tooltip="Mark as read"
+            classname="mt"
+            on:click={() => readNotification(notification.id)}
+          >
+            <svg src="images/icons/check.svg" class="icon" />
+          </Button>
+          <time datetime={notification.timestamp}>
+            {fecha.format(new Date(notification.timestamp), 'HH:mm, DD.MM')}
```
aabounegm: No need for fecha at all. Same thing can be achieved using `new Date().toLocaleString('ru', {
    hour: '2-digit',
    minute: '2-digit',
    day: '2-digit',
    month: '2-digit',
})`

src/components/common/notification-center.svelte
```diff
@@ -46,6 +58,39 @@
...
+          </span>
+          <Button
+            isRound
+            tooltip="Mark as read"
+            classname="mt"
+            on:click={() => readNotification(notification.id)}
+          >
+            <svg src="images/icons/check.svg" class="icon" />
+          </Button>
+          <time datetime={notification.timestamp}>
```
aabounegm: This would look better than the ISO string:
```suggestion
          <time datetime={new Date(notification.timestamp).toLocaleString()}>
```

src/constants/notification-content.js
```diff
@@ -0,0 +1,167 @@
+export default function getNotificationContent({ type, payload }) {
```
aabounegm: Don't see how a function belongs to "constants". Perhaps move it to utils?

src/constants/notification-content.js
```diff
@@ -0,0 +1,167 @@
...
+  let fragments;
+  switch (type) {
+    case "purchase_status_changed": {
```
aabounegm: `case`s do not need curly braces. It is only affecting readability since objects are used a lot inside them

src/constants/notification-content.js
```diff
@@ -0,0 +1,167 @@
...
+    case "manual_transaction": {
+      fragments = [
+        'The administrator',
+      ];
+      if (payload.transaction.change > 0) {
+        fragments.push(`granted ${payload.transaction.change} innopoints to you`);
+      } else {
+        fragments.push(`deducted ${-payload.transaction.change} innopoints from you`);
+      }
+    }
```
aabounegm: I suggest adding a `default` with something like "Unknown notification" so it doesn't fail quietly and we can find the error easily.

src/constants/notification-content.js
```diff
@@ -0,0 +1,167 @@
...
+      fragments = [
+        'Your',
+        {
+          text: (payload.product.type == null ?
+            payload.product.name : `'${payload.product.name}' ${payload.product.type}`),
+          url: '/products/' + payload.product.id,
+        },
+        'purchase',
+        (payload.stock_change.status === 'ready_for_pickup' ?
+          'is ready to be picked up at 319 Office' : 'was rejected'),
```
aabounegm: Spaces missing everywhere. After "your", around "purchase", ...
and in all other notification types

src/components/common/notification-center.svelte
```diff
@@ -46,6 +58,39 @@
...
   {:else}
-    <p>List notifications here</p>
+    <SimplebarList classname="notification-list">
+      {#each notifications as notification (notification.id)}
+        <li class="notification">
+          <span class="content">
+            {#each getNotificationContent(notification) as fragment}
+              {#if typeof fragment === 'string'}
+                {fragment}
+              {:else if 'bold' in fragment}
```
aabounegm: What if it was there but false, somehow?
```suggestion
              {:else if fragment.bold == true}
```

src/components/common/notification-center.svelte
illright: No, this attribute is supposed to contain an ISO string. It's for programmatic parsing
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

src/components/common/notification-center.svelte
illright: Yeah. Previously `fecha` was used in another module which I just refactored, so we might remove this dependency altogether

src/constants/notification-content.js
illright: Well, it specifies the strings of notifications which are more or less constant. I would search for those strings in `constants` first, that's why

src/constants/notification-content.js
illright: Yeah, they don't, but I find it easier to read. This way the blocks are clearly defined

src/constants/notification-content.js
illright: Yeah, this is something I hoped HTML would take care of, but it didn't, so we will have to resort to something else, haven't yet come up with a remedy

src/components/common/notification-center.svelte
illright: We don't care, that's the beauty. This would not only impose additional useless checks, but also constrain the rules needlessly. The only thing we care about is that the `bold` attribute is there, whatever it's value may be. Besides, it will never go beyond this manual usage, do you really want to bother?

src/constants/notification-content.js
illright: I will add the spaces, however, we'll see if that works

src/components/common/notification-center.svelte
aabounegm: It is just that you want the code to be easily extensible in the future. If you ever find the need to modify the fragment to include the attribute (maybe following a strict schema where all fragments are objects with specific attributes, for example), you will meet problems you don't know where they're coming from because you didn't prepare for it. And this example isn't too far fetched if we ever decided we want to use Typescript.
It is just about code design and extensibility in general.

src/components/common/notification-center.svelte
aabounegm: My bad. I thought it was being used as a tooltip.

src/constants/notification-content.js
```diff
@@ -89,25 +89,25 @@ export default function getNotificationContent({ type, payload }) {
...
       break;
     }
 
     case "out_of_stock": {
       fragments = [
-        'The',
+        'The ',
         {
           text: (payload.product.type == null ?
             payload.product.name : `'${payload.product.name}' ${payload.product.type}`),
```
aabounegm: This notification is when a specific variety goes out of stock. Maybe include it in the string?

src/constants/notification-content.js
```diff
@@ -153,13 +153,18 @@ export default function getNotificationContent({ type, payload }) {
...
       if (payload.transaction.change > 0) {
         fragments.push(`granted ${payload.transaction.change} innopoints to you`);
       } else {
         fragments.push(`deducted ${-payload.transaction.change} innopoints from you`);
       }
+      break;
+    }
+
+    default: {
+      fragments = ['Just a message from heaven wishing you a great day'];
```
aabounegm: I was thinking more like "An error occurred, please report it", but that looks nice :joy:

src/components/common/notification-center.svelte
illright: ok, boomer

src/constants/notification-content.js
illright: The crap thing is that there's no simple way to address a variety. Besides, I think it's best to make this notification product-level, not variety-level

src/utils/date-range.js
```diff
@@ -1,26 +1,35 @@
-import fecha from 'fecha';
+const months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
```
aabounegm: No need for this array. Could just use `new Intl.DateTimeFormat('en', {month: 'short'}).format(new Date())`

src/utils/date-range.js
```diff
@@ -1,26 +1,35 @@
...
   if (date.start && date.end) {
-    return parseTwoDates(date);
+    return formatDateRange(date);
   } else if (date.start) {
-    return fecha.format(new Date(date.start), 'YYYY-MM-DD');
+    let dateObj = new Date(date.start);
+    return (
+      dateObj.getFullYear() + '-'
+      + ('' + dateObj.getMonth() + 1).padStart(2, '0') + '-'
+      + ('' + dateObj.getDate() + 1).padStart(2, '0')
```
aabounegm: `new Date().toLocaleDateString('ru')`

src/utils/date-range.js
```diff
@@ -40,18 +49,21 @@ const parseTwoDates = date => {
...
+        } else {
+          // Jun 12 - 15, 2020
+          return `${m1} ${d1} – ${d2}, ${y1}`;
+        }
       } else {
-        // Jun 12 - 15, 2020
-        return `${m1} ${d2} – ${d2}, ${y1}`;
+        // Jun 12 - Jul 15, 2020
+        return `${m1} ${d1} – ${m2} ${d2}, ${y1}`;
       }
```
aabounegm: Check out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat/formatRange

src/constants/notification-content.js
aabounegm: Why so? If size, put in between parentheses. If color, can also be put between parentheses, maybe with a bit of CSS coloring it.
Also, you mean when all varieties of the product go out of stock? So if there is one variety that almost no one wants and just a few of it left, but everything else is out of stock, the admin will still not get notified?

src/utils/date-range.js
illright: Seems like the support is poor

PR #71, opened 2020-01-19
Revive Create Product

illright:
This will do as it says. With unfortunate major refactoring and renaming for consistency with the project terminology.

This branch is ready to be branched out from.


src/components/product/form-variety.svelte
```diff
@@ -0,0 +1,122 @@
...
+    {#if !showSizes}
+      <div class="quantity">
+        <label for="quantity" class="label">Quantity</label>
+        <TextField
+          id="quantity"
+          min="0"
+          placeholder="0"
+          type="number"
+          value={variety.quantity || ''}
+          on:change={(e) => changeField('quantity', e.detail)}
```
aabounegm: Using the event and not the bound value causes it to be a string instead of a number because of how the event is handled in the text-field component
```suggestion
          on:change={(e) => changeField('quantity', parseInt(e.detail, 10))}
```

src/components/product/form-variety.svelte
```diff
@@ -0,0 +1,122 @@
...
+        on:change={(e) => dispatch('new-color', e.target.value)}
+        class="new-color-picker"
+      >
+    </Dropdown>
+
+    {#if !showSizes}
+      <div class="quantity">
+        <label for="quantity" class="label">Quantity</label>
+        <TextField
+          id="quantity"
```
aabounegm: This will cause multiple elements with the same id when creating more than one variety.

src/components/product/preview-card.svelte
```diff
@@ -0,0 +1,105 @@
...
+  );
+  $: colors = product && product.varieties.map(x => x.color).filter(x => x != null);
+  $: sizes = selectedVariety && Object.keys(selectedVariety.sizes);
+  let altText = '';
+  if (product) {
+    altText = 'Product image: ';
+    if (product.type) {
+      altText += `"${product.name}" ${product.type}`;
+    } else {
+      altText += name;
```
aabounegm: I don't see this `name` being defined anywhere
```suggestion
      altText += product.name;
```

src/components/product/image-preview-card.svelte
```diff
@@ -0,0 +1,79 @@
...
+    </div>
+  </div>
+  {#if url}
+    <div class="actions">
+      <svg src="images/icons/drag-handle.svg" class="drag-handle" />
+      <Button
+        isDanger
+        isRound
+        tooltip="Delete image"
+        on:click={(e) => { e.preventDefault(); dispatch('delete-file', { fileIndex: index }); }}
```
aabounegm: I think the Svelte way would be like:
```suggestion
        on:click|preventDefault={(e) => { dispatch('delete-file', { fileIndex: index }); }}
```
though I'm not sure if it will work on elements that are not native HTML

illright: Yep. it won't. Only `once` can be applied to components

aabounegm: So it is about the component and not the event being a DOM event

PR #72, opened 2020-01-23
View product

illright:


illright:
@illright What could speed things up is this review @aabounegm @VanishMax

src/routes/product/[id([0-9]+)]/index.svelte
```diff
@@ -0,0 +1,210 @@
+<script context="module">
```
Comment (VanishMax) [inline src/routes/product/[id([0-9]+)]/index.svelte]: Too much for a page. It's better being split

src/components/ui/snackbar.svelte
```diff
@@ -1,32 +1,36 @@
...
-
-  let snackbar = null;
-  onMount(() => document.body.appendChild(snackbar));
-  onDestroy(() => document.body.removeChild(snackbar));
+  export function open() {
+    isOpen = true;
+    if (timeoutInstance != null) {
+      clearTimeout(timeoutInstance);
+    }
+    timeoutInstance = setTimeout(() => { isOpen = false; }, timeout);
```
aabounegm: Or simply
```suggestion
    timeoutInstance = setTimeout(close, timeout);
```

src/routes/product/[id([0-9]+)]/index.svelte
Comment (illright) [reply src/routes/product/[id([0-9]+)]/index.svelte]: How would you suggest splitting it? I guess we could put the purchase modal in a container, but that's about it

src/components/ui/snackbar.svelte
illright: Nah, the `close` function also involves clearing the timeout, and it's not particularly clear why would we wanna do that, so this one looks clearer to me

PR #86, opened 2020-01-25
Feature/edit product

### LINKED ISSUE #85: Edit Product page

illright:
Closes #85

### DIFF
--- a/src/routes/product/[id([0-9]+)]/edit.svelte
+++ b/src/routes/product/[id([0-9]+)]/edit.svelte
@@ -1,0 +1,352 @@
+<script context="module">
+  import getInitialData from '@/utils/get-initial-data.js';
+  import { groupByColor } from '@/utils/group-varieties.js';
+
+  export async function preload(page, session) {
+    const { unmodifiedProduct, account, colors, sizes } = await getInitialData(this, session, new Map([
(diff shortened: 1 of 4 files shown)


src/routes/product/[id([0-9]+)]/edit.svelte
```diff
@@ -0,0 +1,352 @@
...
+      deletedVarieties: null,
+    };
+
+    for (let field of ['name', 'type', 'description', 'price']) {
+      if (updated[field] !== old[field]) {
+        if (diffs.fields == null) {
+          diffs.fields = { [field]: updated[field] };
+        } else {
+          diffs.fields[field] = updated[field];
+        }
```
Comment (aabounegm) [inline src/routes/product/[id([0-9]+)]/edit.svelte]: Doesn't look so idiomatic. Suggestion:
```js
if (diffs.fields == null)
    diffs.fields = {};
diffs.fields[field] = updated[field];
```

Comment (illright) [reply src/routes/product/[id([0-9]+)]/edit.svelte]: Thought about that, but idk which way really is more idiomatic for JS. The literal way seemed like what I'd opt for in, say, Python, so I went with it here. With your permission, I'll keep it as it is

Comment (aabounegm) [reply src/routes/product/[id([0-9]+)]/edit.svelte]: You may keep it if you don't want to dismiss the approvals, but that style is not so readable

PR #91, opened 2020-02-01
Date and time pickers

illright:
Will implement a calendar for picking a date and a date range, as well as a simple spinner-based time picker. Both can be seen in the UI kit page

### DIFF
--- a/src/components/ui/time-picker.svelte
+++ b/src/components/ui/time-picker.svelte
@@ -4,118 +4,66 @@
   import TextField from './text-field.svelte';
 
   export let value = { hours: null, minutes: null };
-  $: {
-    if (value == null) {
-      value = { hours: null, minutes: null };
-    }
+  if (value == null) {
+    value = { hours: null, minutes: null };
   }
   const minutesInHour = 60;
   const hoursInDay = 24;
 
   const dispatch = createEventDispatcher();
 
-  function incHours() {
-    if (value.hours == null) {
+  function setHours(hours) {
+    if (hours < 0) {
+      value.hours = hoursInDay - 1;
+    } else if (hours >= hoursInDay) {
       value.hours = 0;
     } else {
-      value.hours++;
-      if (value.hours > hoursInDay - 1) {
-        value.hours = 0;
-      }
-    }
-
-    if (value.hours != null && value.minutes != null) {
-      dispatch('change', value);
-    }
-  }
-
-  function decHours() {
-    if (value.hours == null) {
-      value.hours = 0;
-    } else {
-      value.hours--;
-      if (value.hours < 0) {
-        value.hours = hoursInDay - 1;
-      }
-    }
-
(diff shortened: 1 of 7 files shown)


src/components/ui/time-picker.svelte
```diff
@@ -0,0 +1,121 @@
...
+      value.hours--;
+      if (value.hours < 0) {
+        value.hours = hoursInDay - 1;
+      }
+    }
+
+    if (value.hours != null && value.minutes != null) {
+      dispatch('change', value);
+    }
+  }
```
aabounegm: A lot of repetition. Same exact code in `incHours`, except for the `-` instead of `+`. Additionally, `null` is treated as `0` in math operations. I suggest combining into one function with a parameter, and perhaps 2 auxiliary functions for the template section:
```js
function changeHours(offset) {
  value.hours = (value.hours + offset) % hoursInDay;
  if (value.minutes != null)
    dispatch('change', value);
}

const incHours = () => changeHours(1);
const decHours = () => changeHours(-1);
```

src/components/ui/time-picker.svelte
```diff
@@ -0,0 +1,121 @@
...
+      value.minutes = 0;
+    } else {
+      value.minutes--;
+      if (value.minutes < 0) {
+        value.minutes = minutesInHour - 1;
+      }
+    }
+
+    if (value.hours != null && value.minutes != null) {
+      dispatch('change', value);
```
aabounegm: Same comments as for hours

src/components/ui/time-picker.svelte
```diff
@@ -0,0 +1,69 @@
...
+  const dispatch = createEventDispatcher();
+
+  function setHours(hours) {
+    if (hours < 0) {
+      value.hours = hoursInDay - 1;
+    } else if (hours >= hoursInDay) {
+      value.hours = 0;
+    } else {
+      value.hours = hours;
+    }
```
aabounegm: Since it seems like you don't like using modular arithmetic here, I think
```js
    hours = Math.min(hours, hoursInDay - 1);   // to prevent overflow
    hours = Math.max(hours, 0);   // to prevent underflow (negative)
    value.hours = hours;
```
looks more idiomatic.

illright: Idiomatic? Don't think so. Cleaner in terms of code aesthetic, maybe. Expressive, not really. In this case I'd opt for clarity instead of beauty

aabounegm:
Also, I would suggest implementing some kind of click-and-hold on the buttons

illright:
@aabounegm but I applied all the suggestions this time...

illright:
Yeah, click and hold would be cool, but I decided against it for:
- it is not essential
- the value range is not that huge, so unprecise spinning wouldn't amount to much

Maybe some other time

aabounegm:
Yes, but in a different way that is still not so idiomatic

PR #92, opened 2020-02-07
Create production Docker image

aabounegm:


Dockerfile.prod
```diff
@@ -0,0 +1,21 @@
+FROM node:12
```
illright: Maybe we could opt for the latest Node, 13?

Dockerfile.prod
```diff
@@ -0,0 +1,21 @@
...
+
+RUN mkdir -p /app
+WORKDIR /app
+
+# copy the project code
+COPY src /app/src
+COPY static /app/static
+COPY .env .eslintrc package.json package-lock.json rollup.config.js /app/
```
illright: I don't see the `.env` file in Git, is that supposed to be this way?

VanishMax: Don't really see a point to list all the files. Copying `COPY .` could be enough

Dockerfile.prod
```diff
@@ -0,0 +1,21 @@
...
+
+# install our dependencies
+RUN npm install
+
+# Build the project
+RUN npm run build
+
+# expose the server port
+EXPOSE 3000
```
VanishMax: Isn't the port listed in `process.env.PORT`?

Dockerfile.prod
illright: It would copy everything, while the intention was only to copy certain files

Dockerfile.prod
aabounegm: I opted for the latest LTS for stability, which is also what I'm using locally.

Dockerfile.prod
aabounegm: Copying `.env` was the only way to fix the "process is undefined" error in production mode. I couldn't understand why, but it did. Also, `.env` files are not supposed to be checked into version control, but they are required to run the project properly. I was meaning to add this to the README (of the main repo) but forgot.
@VanishMax as Lev pointed out, that would also copy node_modules, .gitignore, Dockerfile, and many more files we don't need

Dockerfile.prod
aabounegm: This line doesn't affect the image at all. It simply acts as documentation that this container should have the port 3000 mapped to something. I can add `ENV PORT 3000` to set it if it wasn't passed in the environment or the .env file, but it seems redundant since it defaults to it anyways

PR #93, opened 2020-02-12
Fix/prepare for production

aabounegm:
Mainly removed redundant env vars and prevented a couple of "Cannot read property * of undefined" errors


aabounegm:
Don't worry. It is already the code running on the VM

PR #95, opened 2020-02-15
Create Project

illright:
\*dragon sounds\*

This is massive and disgusting. Hold tight.

It also works on an outdated version of the backend. Hold tighter if you want to test it on your own machine.  
**upd**: all good now.


illright:
@illright about the review tho...

illright:
Regarding the UX thing, please, add it as an issue. I don't think I currently have the energy to do UX-related enhancements on the page)

PR #97, opened 2020-02-19
Fix/dashboard

aabounegm:
Admin Dashboard should now be working properly


illright:
image
The icon in the empty state should be colored, check your markup

illright:
After I change a purchase's status, then click "edit status", then try to choose another status, the choice doesn't show up to be made. The buttons don't seem to do any meaningful actions, and other entries break down too, for some reason

illright:
Allowing to choose yourself in the individual controls is rather strange. Though it is a way for admins to navigate to their own profile, so I don't complain. Let's leave it as a feature. As for innopoints changing tho...

illright:
The success message after changing innopoints should disappear after a few seconds

aabounegm:
> After I change a purchase's status, then click "edit status", then try to choose another status, the choice doesn't show up to be made. The buttons don't seem to do any meaningful actions, and other entries break down too, for some reason

Cannot reproduce. Was the backend on the last fix branch (or, currently, master)?

> Allowing to choose yourself in the individual controls is rather strange. Though it is a way for admins to navigate to their own profile, so I don't complain. Let's leave it as a feature. As for innopoints changing tho...

So what action do you suggest? It doesn't make sense for an admin to send innopoints or messages to themselves, but will they even try to? And even if they do, what harm does it do? Also, there should probably be a different way of loading their profile, if it is even needed.

illright:
The links for email copying are black, in the markup they are green.

illright:
The time format of the purchases also doesn't match the desired one in markup. There it is DD.MM. Though we could do it HH:MM, DD.MM, if you feel like adding more clarity

src/components/dashboard/purchase-card.svelte
```diff
@@ -3,36 +3,33 @@
   import Button from 'ui/button.svelte';
   import UnclickableChip from 'ui/unclickable-chip.svelte';
   import parseColor from '@/utils/optimal-color';
```
illright: `parseColor` doesn't really reflect what the function is doing. I imported it under the name `getBackground` elsewhere, maybe it's worth changing. And while you're at it, append a `.js` suffix to the import path for consistency.

src/components/dashboard/purchase-card.svelte
```diff
@@ -3,36 +3,33 @@
...
     </div>
     <div class="product-info">
       <div class="title">
-        {purchase.name}
-        <span class="type">{purchase.type}</span>
+        {purchase.product.name}
+        <span class="type">{purchase.product.type || ''}</span>
       </div>
-      <time>{purchase.date}</time>
+      <time>{new Date(purchase.time).toLocaleString()}</time>
```
illright: I wrote an earlier comment about the date format, will start a conversation to allow to resolve.

src/components/dashboard/send-comment.svelte
```diff
@@ -1,18 +1,31 @@
 <script>
   import TextField from 'ui/text-field.svelte';
   import Button from 'ui/button.svelte';
+  import { post } from '@/utils/api';
```
illright: Same note on import suffixes

src/components/dashboard/send-comment.svelte
```diff
@@ -1,18 +1,31 @@
...
   let success = false;
   let text = null;
+  let error = null;
 
-  const send = () => {
-    text = null;
-    success = true;
-  };
+  async function send() {
+    if(!text) return;
```
illright: also i'd love to see `if` being separated from the `(` with a space :p
same thing on line 19

src/components/ui/autocomplete.svelte
```diff
@@ -34,7 +34,7 @@
   let optionGen;
   let promises = [];
   let hasMoreOptions;
-  let selected = new Set(values.map(opt => opt.name));
```
illright: Nah, I don't approve of those changes. This is a UI component, the data format shouldn't have to imply that it is only used for users with a full name and the extra data is always an email. That's why in the Create Project I've simply rewired the objects coming from the backend before passing them over to the Autocomplete component.

src/components/ui/dropdown.svelte
```diff
@@ -24,6 +24,7 @@
 
   let dropdownShell = null;
   const clickOutside = (event) => {
+    if(!dropdownShell) return;
```
illright: code style pls

src/constants/autocomplete-options.js
```diff
@@ -1,45 +1,20 @@
-function sleep(ms) {
-  return new Promise(resolve => setTimeout(resolve, ms));
-}
+import { get } from '@/utils/api.js';
```
illright: Okay, and this file was supposed to remain constant :) It feels like we have a certain level of misunderstanding as to what should go in `utils/` and what should go to `constants/`. I'd rather revert this file for the sake of the UI page which uses it in constant format and move this functionality to `utils/`. Anyway, we should probably talk this over in person

src/routes/product/[id([0-9]+)]/edit.svelte
```diff
@@ -295,7 +295,7 @@
   <link rel="stylesheet" href="css/page-components/footer.css" />
 </svelte:head>
 
-<Layout>
+<Layout user={account}>
```
Comment (illright) [inline src/routes/product/[id([0-9]+)]/edit.svelte]: So if you moved away from the `$session.user`, I suppose there's no longer a reason to set it on every page.

Comment (illright) [reply src/routes/product/[id([0-9]+)]/edit.svelte]: To be precise, I'm talking about lines like https://github.com/VanishMax/innopoints-frontend/blob/a994c3521f94b81842a4d2e8c77df155fd5aa8be/src/routes/profile.svelte#L48 which I used to write before

src/routes/dashboard.svelte
```diff
@@ -16,9 +33,9 @@
   <link rel="stylesheet" href="/css/page-components/empty-state.css" />
 </svelte:head>
 
-<Layout title="Dashboard" adminsOnly>
+<Layout user={account}>
   <h1 class="padded">Dashboard</h1>
-  <div class="cards padded">
+  <div class="cards padded material">
```
illright: `div.material` should be wrapping `h1.padded` and `div.cards.padded`, as per the markup

src/containers/dashboard/purchases.svelte
```diff
@@ -2,11 +2,22 @@
...
-  const reject = e => purchases = purchases.filter(x => x.name !== e.detail.name);
-  const deliver = e => reject(e);
+  async function setStatus(stock_change, status) {
+    await patch(`/stock_changes/${stock_change.id}/status`, {
+      data: {
+        status,
+      },
+    });
+    stock_change.status = status;
+    if([StockChangeStatuses.CARRIED_OUT, StockChangeStatuses.REJECTED].includes(status)) {
```
illright: `if (`

src/containers/dashboard/purchases.svelte
```diff
@@ -2,11 +2,22 @@
...
   import Button from 'ui/button.svelte';
   import PurchaseCard from '@/components/dashboard/purchase-card.svelte';
+  import { patch } from '@/utils/api.js';
+  import StockChangeStatuses from '@/constants/backend/stock-change-statuses.js';
 
   export let purchases = [];
-  // In real-case situation would be a bit different
-  const reject = e => purchases = purchases.filter(x => x.name !== e.detail.name);
-  const deliver = e => reject(e);
+  async function setStatus(stock_change, status) {
```
illright: it's rather strange that you went with `snake_case` for the `stock_change` variable. Not like it's coming from the backend or anything

src/components/ui/autocomplete.svelte
aabounegm: So do you suggest to change the function in "@/constants/autocomplete-options.js" or the wiring between importing the function and passing it to the autocomplete component?

src/routes/product/[id([0-9]+)]/edit.svelte
Comment (aabounegm) [reply src/routes/product/[id([0-9]+)]/edit.svelte]: Should I just remove this or all of them from everywhere? Note that there might be some components that still make use of it without accepting it as a prop yet, but that should also be fixed and isn't too difficult

src/containers/dashboard/purchases.svelte
aabounegm: Was just consulting the backend for the payload and so on, so it was an unconscious choice.

src/routes/product/[id([0-9]+)]/edit.svelte
Comment (aabounegm) [reply src/routes/product/[id([0-9]+)]/edit.svelte]: Actually, nvm. It took 2 minutes to remove them all

src/components/dashboard/purchase-card.svelte
```diff
@@ -30,7 +30,7 @@
       <time>{new Date(purchase.time).toLocaleString()}</time>
       <div class="purchaser">
         {(-purchase.amount)} item{purchase.amount < -1 && 's' || ''} purchased by {purchase.account.full_name}
-        <span on:click={copy} class="copy-email">click to copy e-mail</span>
+        <a on:click={copy} href="# " class="copy-email">click to copy e-mail</a>
```
illright: Yeah, it's not really conventional to use # in Sapper routes, mainly for the reason that this link will now bring the user to `/# ` as opposed to `/dashboard# `. I guess we could use a span, just give it the styles of a link:
```scss
text-decoration: underline;
color: $main;
cursor: pointer;
```

src/utils/list-users.js
```diff
@@ -10,10 +10,10 @@ export default async function *getOptions(query) {
 
   do {
     const res = await get(`/accounts?page=${current_page}&limit=${limit}&q=${query}`);
-    const accounts = await res.json();
-    yield accounts.data;
+	const {data: accounts, pages: new_pages} = await res.json();
```
illright: Indentation messed up. Also, put spaces around curly braces in destructuring, like `{ data: accounts, pages: newPages }` (yeah, and casing)

PR #110, opened 2020-02-26
Feature/admin view profile

aabounegm:
Allows admins to see profiles of other users.

### DIFF
--- a/src/routes/profile.svelte
+++ b/src/routes/profile.svelte
@@ -3,21 +3,39 @@
   import isoForURL from '@/utils/iso-for-url.js';
 
   export async function preload(page, session) {
+    let requestedEmail = page.query.user;
     let timelineFetchedUntil = new Date();
     timelineFetchedUntil.setMonth(timelineFetchedUntil.getMonth() - 3);
+    const { currentUser } = await getInitialData(this, session, new Map([
+      ['currentUser', '/account'],
+    ]));
+    if (currentUser == null) {
+      this.error(403, 'Profile');
+    }
+    const isMe = requestedEmail == currentUser.email;
+    const allowed = requestedEmail == null || isMe || currentUser.is_admin;
+    // non-admin requested a different profile
+    if (!allowed) {
+      this.redirect(302, `/profile`);
+    }
+    if (requestedEmail == null) {
+      requestedEmail = currentUser.email;
+    }
+
     const {
       account,
       ...initialData
     } = await getInitialData(this, session, new Map([
-      ['account', '/account'],
-      ['timeline', `/account/timeline?start_date=${isoForURL(timelineFetchedUntil)}`],
-      ['statistics', `/account/statistics?start_date=${isoForURL(timelineFetchedUntil)}`],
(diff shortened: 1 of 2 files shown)


src/routes/profile.svelte
```diff
@@ -3,21 +3,39 @@
...
+
     const {
       account,
       ...initialData
     } = await getInitialData(this, session, new Map([
-      ['account', '/account'],
-      ['timeline', `/account/timeline?start_date=${isoForURL(timelineFetchedUntil)}`],
-      ['statistics', `/account/statistics?start_date=${isoForURL(timelineFetchedUntil)}`],
-      ['notificationSettings', '/account/notification_settings'],
+      ['account', `/accounts/${requestedEmail}`],
```
illright: Why are you requesting /account for the second time? (first at line 9)

src/routes/profile.svelte
```diff
@@ -3,21 +3,39 @@
...
+      ['currentUser', '/account'],
+    ]));
+    if (currentUser == null) {
+      this.error(403, 'Profile');
+    }
+    const isMe = requestedEmail == currentUser.email;
+    const allowed = requestedEmail == null || isMe || currentUser.is_admin;
+    // non-admin requested a different profile
+    if (!allowed) {
+      this.redirect(302, `/profile`);
```
illright: I think a 403 would work better. This is an access violation, and is not something that will naturally occur on the website, so it makes sense to report an error instead of silently redirecting to the allowed resource

src/routes/profile.svelte
```diff
@@ -54,7 +72,7 @@
   }
 
   function changeUsername({ detail }) {
-    api.patch('/account/telegram', {
+    api.patch(`/account/${account.email}/telegram`, {
```
illright: Shouldn't it be `/accounts` instead of `/account`? Same thing for subsequent changes

src/routes/profile.svelte
aabounegm: They don't have to be the same account. The one requested here is of the profile we are viewing rather than the logged-in user. I understand that it is redundant in cases where a user is viewing their own profile, but the code would be a bit messier to handle that case.

src/routes/profile.svelte
aabounegm: Quoting your messages on Telegram:
> Lev Chelyadinov, [16.02.20 08:49]
> Like, if they ever get this url

> Lev Chelyadinov, [16.02.20 08:49]
> They should be thrown to their own profile

> Lev Chelyadinov, [16.02.20 08:49]
> Instead of 403

You argued that it should be a query param instead of a route one for exactly this reason, to simply redirect by removing the query in case it wasn't the user's own profile

src/routes/profile.svelte
aabounegm: Fixed in a commit I cannot see here for some reason

src/routes/profile.svelte
illright: Hm. What was I thinking. Okay then, let's leave it as it is :D

PR #118, opened 2020-03-18
Hotfix/product without image

### LINKED ISSUE #100: The View Product page crashes for products without images
https://github.com/VanishMax/innopoints-frontend/blob/42a8cdf415f9e4fe746c2dd605218878d4053ae6/src/containers/product/image-previews.svelte#L11

The above line will fail if the product is completely imageless.

aabounegm:
Fixes #100

### DIFF
--- a/src/containers/products/image-previews.svelte
+++ b/src/containers/products/image-previews.svelte
@@ -9,7 +9,7 @@
   export let productControl;
   export let selectedColor;
 
-  if(productControl.flatImages.length === 0) {
+  if (productControl.flatImages.length === 0) {
     productControl.flatImages.push({
       url: '/images/create-product/placeholder.svg',
       placeholder: true,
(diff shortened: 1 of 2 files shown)


src/containers/products/image-previews.svelte
```diff
@@ -8,6 +8,13 @@
 
   export let productControl;
   export let selectedColor;
+
+  if(productControl.flatImages.length === 0) {
```
illright: Put a space here, please :)
Wonder if there's an entry to ESLint to enforce this layout

aabounegm: I don't know why I keep missing this particular one that much.
Yes, there is (https://eslint.org/docs/rules/keyword-spacing), and the default is to check for this. Why it is not reporting anything is beyond me.

aabounegm: Or on another look, I may have simply overlooked it previously because the console gets flooded with nginx logs right after the change due to hot reloading.
I added the rule anyways

PR #120, opened 2020-03-18
Hotfix/values keyword

aabounegm:
For some weird reason (that took hours to find out its cause), using the prop name `values` causes the following error:
image
It has been observed only when running in Docker in development mode, but it is still worth fixing. The root cause of this problem is unknown (couldn't find anything on the internet), but this workaround fixes it anyway.
All `values` props have been renamed to `items`.

### DIFF
--- a/src/containers/projects/new/step-3.svelte
+++ b/src/containers/projects/new/step-3.svelte
@@ -88,7 +88,7 @@
     You are the creator, therefore, also a moderator.
   </p>
   <span class="label">Search for people</span>
-  <Autocomplete getOptions={getUsers} on:change={recordChanges} />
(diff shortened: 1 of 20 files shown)


src/containers/projects/new/step-3.svelte
```diff
@@ -88,7 +88,7 @@
     You are the creator, therefore, also a moderator.
   </p>
   <span class="label">Search for people</span>
-  <Autocomplete values={values} getOptions={getUsers} on:change={recordChanges} />
```
illright: I think this should've been replaced for `selection`, because this prop is a way to set the initial selection of values, which will be needed for, say, editing the project.

aabounegm: yeah, I didn't really know what it referred to and assumed it was before `getOptions`

PR #124, opened 2020-03-24
Hotfix/purchase status notification

aabounegm:
Handles all stock_change status types in the notification.
Changes user hyperlink to the user's profile instead of sending an email

### DIFF
--- a/src/constants/notification-content.js
+++ b/src/constants/notification-content.js
@@ -1,12 +1,14 @@
+import StockChangeStatuses from '@/constants/backend/stock-change-statuses';
+
 export default function getNotificationContent({ type, payload }) {
   let fragments;
   switch (type) {
     case "purchase_status_changed": {
       const statuses = {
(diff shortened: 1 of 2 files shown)


src/constants/notification-content.js
```diff
@@ -10,8 +16,7 @@ export default function getNotificationContent({ type, payload }) {
           url: '/products/' + payload.product.id,
         },
         ' purchase ',
-        (payload.stock_change.status === 'ready_for_pickup' ?
```
illright: lmao this is just great

src/constants/notification-content.js
```diff
@@ -2,6 +2,12 @@ export default function getNotificationContent({ type, payload }) {
   let fragments;
   switch (type) {
     case "purchase_status_changed": {
+      const statuses = {
```
illright: Consider using [stock-change-statuses.js](https://github.com/VanishMax/innopoints-frontend/blob/develop/src/constants/backend/stock-change-statuses.js)

PR #127, opened 2020-03-24
Hotfix/purchase dialog

aabounegm:


src/containers/products/item-content.svelte
```diff
@@ -20,16 +20,19 @@
   let animation = false;
 
   function purchase() {
-    if (!productControl.productSized || selectedVarietyID != null) {
-      err = false;
-      dispatch('purchase', { varietyID: selectedVarietyID });
-    } else {
+    // not in sizes, select variety automatically
+    if (!productControl.productSized) {
+      const selectedVariety = productControl.varietiesByColor.get(selectedColor);
```
illright: This could be confusing, since from the point of view of code, a variety is a pair (color, size), while from the point of view of the user a variety is categorized only by color, and a variety has different sizes. I suggest naming this variable to reflect the sense that this is actually a selected color which acts as a grouping parameter for varieties

src/routes/products/new.svelte
```diff
@@ -115,75 +115,69 @@
     persistToStorage();
   }
 
-  function createProduct() {
+  async function createProduct() {
     prefetch('/store');
     if (!product.name) {
       errors.name = true;
       errorMessage = 'Some fields are not filled out or filled out incorrectly.';
+      return;
```
illright: Perhaps it would be better to omit these returns to have all validation errors visible at the same time

src/containers/products/item-content.svelte
aabounegm: What do you suggest? I didn't quite understand that map object

illright: I suggest renaming the `selectedVariety` variable to anything that would reflect that this is in fact an array of varieties

PR #136, opened 2020-03-28
View project

### LINKED ISSUE #94: Discarded changes somehow persist when editing activities
Steps to reproduce:
- go to Create Project
- open a draft that has activities
- click "edit" on one of the activities
- change the dates of the activity to something else
- discard the changes
- cancel the editing
- click "edit" on that activity again

The changes you previously discarded will reappear.

illright:
\*harder dragon sounds\*
Also features the Review Project page.
Closes #94


src/components/projects/new/competence-picker.svelte
```diff
@@ -0,0 +1,39 @@
...
+    dispatch('change', value);
+  }
+
+  const dispatch = createEventDispatcher();
+</script>
+
+<Dropdown label="select competences" dropdownclass="btn-shift" noclose nowrap>
+  <CheckboxGroup
+    name="competences{activityIndex}"
+    maxChecked={3}
```
aabounegm: I suggest parametrizing this value to allow for reusability when it comes to giving feedback.
Not necessarily right now, though. Just remember it when it comes to the other feature

src/components/projects/view/danger-confirm-dialog.svelte
```diff
@@ -0,0 +1,33 @@
...
+    <svg slot="icon" src="images/icons/alert-triangle.svg" class="icon" />
+    <p slot="content">
+      <slot />
+    </p>
+    <div slot="content" class="actions">
+      <Button on:click={() => { isOpen = false; dispatch('reject', eventDetail); }}>
+        no, wait
+      </Button>
+      <Button isDanger isFilled on:click={() => dispatch('confirm', eventDetail)}>
+        {textYes}
```
aabounegm: Why is this text a parameter (with no default) when the "no text" is not?

src/components/projects/view/leave-feedback-modal.svelte
```diff
@@ -0,0 +1,87 @@
...
+            <label for="competences">
+              What competences have you developed (up to 3)?
+              <span class="required">*</span>
+            </label>
+            <CheckChipGroup
+              classname="competence-chips"
+              items={competenceItems}
+              name="competences"
+              labels={competenceLabels}
+              maxChecked={3}
```
aabounegm: When leaving feedback, Andrey mentioned that students can select more than 3

src/components/projects/view/moderator-activity-card.svelte
```diff
@@ -0,0 +1,160 @@
...
+  import ApplicationStatuses from '@/constants/backend/application-statuses.js';
+  import s from '@/utils/plural-s.js';
+  import { formatDate, formatTimeRange } from '@/utils/date-time-format.js';
+
+  export let activity;
+  export let competences;
+
+  $: pendingApplications = (
+    activity.applications == null ?
+    [] : activity.applications.filter(x => x.status === ApplicationStatuses.PENDING)
```
aabounegm: Convention when using ternary operator on multiple lines appears to be something like this:
```suggestion
    activity.applications == null
      ? []
      : activity.applications.filter(x => x.status === ApplicationStatuses.PENDING)
```

src/containers/projects/view/project-header.svelte
```diff
@@ -0,0 +1,149 @@
...
+  export let project;
+  export let account;
+  export let review = false;
+
+  let reviewComment = null;
+
+  const projectImageURL = (
+    project.image_id == null ?
+      '/images/create-project/placeholder.svg'
+    : `${API_HOST}/file/${project.image_id}`
```
aabounegm: Same thing here
```suggestion
    project.image_id == null
      ? '/images/create-project/placeholder.svg'
      : `${API_HOST}/file/${project.image_id}`
```

src/containers/projects/view/user-view.svelte
```diff
@@ -0,0 +1,25 @@
+<script>
+  import UserActivityCard from '@/components/projects/view/user-activity-card.svelte';
+
+  export let projectStage;
+  export let activities;
+  export let competences;
+  export let account;
+
+  $: activityCards = activities.filter(x => !x.internal);
```
aabounegm: Why does it need to be reactive?

src/routes/projects/index.svelte
```diff
@@ -39,6 +39,7 @@
       queryArgs.set('spots', filtering.vacantSpots.toString());
     }
 
+    // TODO: implement
```
aabounegm: and why is it not implemented?

src/utils/project-manipulation.js
```diff
@@ -32,12 +38,155 @@ export function filterProjectFields(project, edit = false) {
...
+        internalListIdx++;
+        continue;
+      }
+
+      // If the internalList cursor is on a displayed activity,
+      // don't change anything – it is already that same activity
+      if (internalList[internalListIdx]._type === ActivityTypes.DISPLAY) {
+        internalListIdx++;
+        activityProcessed = true;
+        break;
```
aabounegm: Either the variable or the `break;` statements are useless

src/components/projects/view/moderator-activity-card.svelte
illright: Idk, I find it excessive, besides, this way kind of keeps the `?` at the line of the condition, which makes more sense to me

src/routes/projects/index.svelte
illright: This is out of scope for this PR)

src/utils/project-manipulation.js
illright: Yes, the variable was introduced for the readability of this while loop. Removing it would make the loop `while (true)`, and it's not clear what's the terminating condition for that loop. Just done to preserve sanity of future coders

src/components/projects/view/moderator-activity-card.svelte
aabounegm: I think it improves readability, making it look more similar to `if ... else` (by giving each branch a different line) and aligning both lines by starting each with a symbol (`?` and `:` respectively)

src/containers/projects/view/project-header.svelte
illright: Same response

src/utils/project-manipulation.js
aabounegm: So that makes the `break;` unnecessary

src/components/projects/view/leave-feedback-modal.svelte
illright: That's bullshit, because he always changes his mind. Let's keep track of a list of explicit questions we have for him, and ask them all

src/components/projects/view/danger-confirm-dialog.svelte
illright: Not striving for 100% reusability here, this is a page-specific component. As for the default, same thing, there's always a value in our use cases

src/components/projects/new/competence-picker.svelte
illright: Don't get what you mean

src/utils/project-manipulation.js
illright: No, the `break` ensures that the loop execution doesn't go further.

src/components/projects/view/moderator-activity-card.svelte
illright: We might have to agree to disagree on this

src/components/projects/new/competence-picker.svelte
aabounegm: When volunteers submit feedback, they are allowed to choose more than 3 competences. My suggestion is to move this number into a prop so that the same component can be used when submitting feedback (if it would be useful there). If the interface itself is different then nvm

src/utils/project-manipulation.js
aabounegm: and why do you need that? It seems that all if-conditions are mutually exclusive anyways. They are all checking if one property is `===` something

illright: The `internalListIdx` gets modified in process, which means these conditions wouldn't be checking the same thing. So they are not mutually exclusive in that sense

src/components/projects/new/competence-picker.svelte
illright: It is different. In the other picker is a set of chips

src/containers/projects/view/user-view.svelte
illright: Not really sure, wish I could give a more precise answer to this. Maybe it's to make the interface react to changes in the activities from the outside.
Short answer: it works like this, and I wouldn't really want to go seeking different solutions right now) you can mark this spot though, perhaps we could optimize it later

aabounegm: I wouldn't really call it "different solutions", it is just removing 2 symbols that add a lot of overhead in observers. wouldn't take a minute to make it works as it should.
I don't think activities are supposed to "change from the outside", especially in the user view; at least not without reloading the whole page.
If you still fear the lack of reactivity, you can just put this expression directly in the template in place of the `activityCards` variable; it is short enough.

illright: That's the problem, there's a lot more happening with the internal variables than you might imagine, that's why I wouldn't be so quick to make conclusions. I understand the overhead, it's just that, as of now, fixing it is dangerous. Note that most of the requests to the API on this page update the project object on their own, without rerequesting it from the backend

PR #139, opened 2020-03-28
Fix minor issues

illright:
Best to review these commit-by-commit, since they are independent and pretty small


src/routes/products/[id([0-9]+)]/edit.svelte
```diff
@@ -182,7 +182,7 @@
   }
 
   function saveChanges() {
-    prefetch('/store');
+    prefetch(`/products/${product.id}`);
```
Comment (aabounegm) [inline src/routes/products/[id([0-9]+)]/edit.svelte]: Doesn't seem appropriate to prefetch the product before the changes are applied and saved.
On a side note, this whole function can be refactored a lot

PR #146, opened 2020-03-29
CSS minification

illright:


src/routes/_error.svelte
```diff
@@ -17,10 +17,7 @@
   {:else}
     <title>Something went wrong...</title>
   {/if}
-  <link rel="stylesheet" href="/css/page-components/header.css">
-  <link rel="stylesheet" href="/css/page-components/404.css">
-  <link rel="stylesheet" href="/css/page-components/403.css">
-  <link rel="stylesheet" href="/css/page-components/footer.css">
+  <link rel="stylesheet" href="/css/bundles/-error.min.css" />
```
VanishMax: What is this dash?

illright: Well, the page is called `_error`, and in the script I replace all `_` with `-`, figured this is not a problem, so I kept it

src/routes/index.svelte
```diff
@@ -27,16 +27,16 @@
...
+  <link rel="stylesheet" href="/css/bundles/home.min.css" />
+  <link rel="preload" as="style" href="/css/bundles/store.min.css" />
+  <link rel="preload" as="style" href="/css/bundles/projects.min.css" />
+  {#if account}
+    {#if account.is_admin}
+      <link rel="preload" as="style" href="/css/bundles/dashboard.min.css" />
+    {:else}
+      <link rel="preload" as="style" href="/css/bundles/profile.min.css" />
+    {/if}
+  {/if}
```
aabounegm: Here you want to use prefetch instead of preload
Check out https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content (look at the **Other resource preloading mechanisms** section)

PR #170, opened 2020-04-04
Fixing bugs

illright:
Most of the fixes are individual, to be reviewed on a per-commit basis.


src/components/dashboard/purchase-card.svelte
```diff
@@ -24,13 +25,12 @@
         {purchase.product.name}
         <span class="type">{purchase.product.type || ''}</span>
       </div>
-      <time datetime={purchase.time} title={new Date(purchase.time).toLocaleString('ru')}>
-        {new Date(purchase.time).toLocaleString('ru', {month: '2-digit', day: '2-digit'})}
-      </time>
+      <time datetime={purchase.time}>{formatTime(purchase.time)}</time>
```
aabounegm, line 28: Why did you remove the `title` attribute? It can be helpful to know the exact time when hovering over it

src/components/projects/view/application-dialog.svelte
```diff
@@ -22,6 +22,10 @@
       remember: remember[0].checked,
     });
     comment = '';
+    if (!remember[0].checked) {
+      telegram = savedUsername || '';
+    }
+    remember[0].checked = false;
```
aabounegm, line 28: Why forget whether or not they want to **remember**?)

src/components/projects/view/application-hour-tile.svelte
```diff
@@ -9,12 +9,19 @@
...
   export let application;
   export let handlePanelOpen;
+  $: report = application.reports.find(report => report.reporter_email === account.email);
 
   const dispatch = createEventDispatcher();
 </script>
 
 <AccordionSection on:panel-open={handlePanelOpen} let:toggle>
-  <button slot="handle" class="btn handle round" type="button" on:click={toggle}>
+  <button
```
aabounegm, line 18: Why is it lowercase? Does that still make it our imported component or the native HTML element?

src/containers/store/tagline.svelte
```diff
@@ -1,8 +1,23 @@
...
   import Dot from 'ui/dot.svelte';
+  import * as api from '@/utils/api.js';
 
   export let isAdmin;
+  let dotShown = false;
+  
+  onMount(async () => {
+    if (isAdmin) {
+      try {
+        const purchases = await api.json(api.get('/stock_changes/for_review'));
```
aabounegm, line 13: Perhaps it's worth adding a `?limit=1` param (to backend first) so it doesn't request a lot of unnecessary information since it doesn't store it in a central store or anything and will re-request it on the other page.
But that's probably out of the scope of this PR

src/components/dashboard/purchase-card.svelte
illright: This representation is already as exact as it gets (sans seconds, but who gives a shit)

src/components/projects/view/application-dialog.svelte
illright: If they remembered it once, no need to send subsequent saving requests, it's already saved

src/components/projects/view/application-hour-tile.svelte
illright: Native. The difference is that this one doesn't have a ripple attached, which was messing things up

src/containers/store/tagline.svelte
illright: Yeah, thought about that too, but brushed it off thinking there will never be a surplus of purchases, and since this is asynchronous, it won't slow anything down

src/components/projects/view/application-hour-tile.svelte
illright: Or, actually, no. The reason why the native element is used here is because of the slot. But there's one more place in code where a native button is preferred for ripple reasons

src/components/projects/view/application-dialog.svelte
aabounegm: But is that obvious to the user, or will they click "remember" on every application?

src/components/dashboard/purchase-card.svelte
aabounegm: ah, ok. Previously, it was only month and day, but I didn't check `formatTime` after this change

src/components/projects/view/application-dialog.svelte
illright: It is, since the field will be pre-filled on subsequent applications (and not on the first one)

PR #171, opened 2020-04-04
Prevent backend overflow by limiting user input

illright:


src/components/dashboard/send-money.svelte
```diff
@@ -42,6 +42,7 @@
       type="number"
       isWithItem
       isItemRight
+      maxlength={8}
```
aabounegm, line 45: it has `type="number"`. Shouldn't this be something like `max={99999999}` instead?

illright: Doesn't really matter

PR #179, opened 2020-04-10
Feature/templates

illright:


illright:
> Did you review the templates with Andrey?
That was his templates file, so yeah. As for the other templates, I told him that it would come in a later stage
> Perhaps having to manually delete them would be safer
There are many issues related to drafts and templates, I must confess that the system is pretty fragile and poorly constructed, but I tried to make the best of it. There are strict requirements on what data should the activities contain before they can be saved to the backend even as a draft, so that's what motivated the current intended behaviour

PR #185, opened 2020-04-18
Feature/image crop

illright:
Talk about technical debt. This is ugly from many perspectives, but should be good enough to last us a few more usability tests until the grande rewrite some time in the future


illright:
1. Weird, I thought scrolling was pretty intuitive for zoom. For now let's merge it like this, later, when we work on the UX, we'll add it
2. It shouldn't, kinda, but detecting image dimensions in JS is a pain in the ass, so I figured showing the dialog wouldn't hurt. When the entire area is selected, the backend bypasses the cropping
3. Again, this is ugly and temporary. We just needed something that would prevent the majority of users from breaking our grid, the product creation is less of a priority. Also, it supports several images, so it's a bit more complex in that regard. Just bear with it for now, everything will come

aabounegm:
1- I was using the touchpad and not a mouse, so it wasn't the first thing that came to mind. Perhaps add an issue to remember it?
2- Doesn't seem so difficult (https://stackoverflow.com/a/47786555), although a bit inelegant, but fine, we can do it later. Perhaps an issue to remind would be great.

PR #188, opened 2020-04-19
Update wording for clarity

illright:


illright:
One will need another PR to develop to update the bundles, since I didn't resolve the conflict properly

PR #220, opened 2020-04-22
Telegram handling

illright:
Idk how could we have missed that, but previously one could save a wrong Telegram value


aabounegm:
Then shouldn't the same thing be done in the application dialog as well?

illright:
Oh, definitely. Will fix

illright:
Had to do some gool ol' boolean logic on a piece of paper for this commit xD

Issue #217, opened 2020-04-21
The time picker is unpleasant to use


illright:
To fill out both fields you need to click through both fields.

Proposal:
- [ ] Default the picker to 23:59 (seems like the most reasonable deadline time)
- [ ] Switch the focus to the second field when the first field has received a valid value

aabounegm:
The second point would require modification to the TextField component itself since it needs reference to the HTML element to access events (such as `blur`) and methods (such as `focus()`), which seems like too much work to do given that the outcome is just saving one "tab" button press.

PR #221, opened 2020-04-24
Hotifx/time picker

aabounegm:
Relates to #217. Can close it depending on the discussion there.

### DIFF
--- a/src/components/projects/new/edit-activity.svelte
+++ b/src/components/projects/new/edit-activity.svelte
@@ -327,6 +327,11 @@
             toggle(e);
             fieldsAltered = true;
           }}
+          on:change-date={({detail}) => {
+            if (activity.application_deadline != null) return;
+            activity.application_deadline = detail;
+            activity.application_deadline.setHours(23, 59);
+          }}
         />
       </Dropdown>
     </FormField>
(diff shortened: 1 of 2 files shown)


src/components/projects/new/edit-activity.svelte
```diff
@@ -327,6 +327,11 @@
             toggle(e);
             fieldsAltered = true;
           }}
+          on:change-date={({detail}) => {
```
illright, line 330: Put spaces around the key in the object, for consistency with the rest of the codebase

src/components/ui/date-time-picker.svelte
```diff
@@ -46,11 +43,25 @@
...
+      dispatch('change-date', dateValue);
+      setTimeout(() => datePickerShown = false, 10);
+    }} />
   {:else}
     <div class="time-wrapper">
       <svg class="icon" src="images/icons/clock.svg" />
-      <TimePicker bind:value={timeValue} />
+      <TimePicker
+        value={timeValue}
+        on:change={({detail}) => {
```
illright, line 59: Here too

illright:
The thing is, people don't use the Tab key to switch through fields. That's why the second point was somewhat important. It is less so now that the time is already set to be reasonable, because people will have less reasons to alter it, but I still would keep the issue open. Even though the next time it will be touched will probably be after the migration to Attractions, which would have this feature built-in

PR #222, opened 2020-04-24
Disable the button while uploading image

### LINKED ISSUE #216: The Uploading indicator of the resizer popup
Right now, when you try to upload an image and you're faced with a resizer dialog, clicking the "upload" button provides any visible results only after a while, tempting users to click the button once again. We should replace the button with a loading indicator, perhaps.

aabounegm:
Closes #216

### DIFF
--- a/src/components/common/image-resizer.svelte
+++ b/src/components/common/image-resizer.svelte
@@ -67,7 +67,7 @@
           }}
           disabled={uploading}
         >
-          { uploading ? 'uploading...' : 'upload' }
+          {uploading ? 'uploading...' : 'upload'}
         </Button>
       </div>
     </Dialog>
--- a/src/components/common/image-resizer.svelte
+++ b/src/components/common/image-resizer.svelte
@@ -11,7 +11,7 @@
   export let aspectRatio = 1;
   export let isOpen = false;
   export let error = null;
-  export let uploading = false;
+  export let isUploading = false;
 
   function saveBounds(event) {
     pixels = event.detail.pixels;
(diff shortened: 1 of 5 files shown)


aabounegm:
or perhaps a simpler option c) disable the "upload cover" button while it is uploading

illright:
Yeah, decent, actually. Change the text too

src/components/common/image-resizer.svelte
```diff
@@ -56,12 +57,17 @@
...
-          on:click={() => dispatch('image-cropped', pixels)}
+          on:click={() => {
+            uploading = true;
+            dispatch('uploading', uploading);
+            dispatch('image-cropped', pixels);
+          }}
+          disabled={uploading}
         >
-          upload
+          { uploading ? 'uploading...' : 'upload' }
```
illright, line 70: As for these braces, I wasn't putting spaces around them :) Only objects and code blocks, not Svelte JS embeddings

src/components/projects/view/moderator-hour-card.svelte
```diff
@@ -45,11 +45,12 @@
         />
         {#if enteredHours !== application.actual_hours}
           <Button
-            isRound
+            isFilled
             disabled={enteredHours < 0}
             on:click={dispatchHourChange}
           >
-            <svg class="icon" src="images/icons/check.svg" />
+            <svg class="icon" src="images/icons/save.svg" />
```
illright, line 52: This should probably be `class="icon mr"` then

src/components/projects/view/moderator-hour-card.svelte
```diff
@@ -45,11 +45,12 @@
...
+            save hours
```
illright, line 53: Let's keep the text to just "save", it's too long otherwise

src/containers/projects/new/step-1.svelte
```diff
@@ -52,7 +53,7 @@
       <span class="lb">This is shown on the project card to catch attention.</span>
       <span class="lb">Best to use 16:9 photos.</span>
     </span>
-    <ProjectImagePicker bind:value={$project.image_id} on:resize-image />
+    <ProjectImagePicker bind:value={$project.image_id} {uploading} on:uploading on:resize-image />
```
illright, line 56: Ugh, the event and the prop have the same name. If you could change one of them, I'd appreciate.

src/components/projects/view/moderator-hour-card.svelte
aabounegm: I removed it on purpose as it was too long, but I will put it back since I will remove the word "hours"

PR #223, opened 2020-04-24
Hotfix/empty reports dialog

### LINKED ISSUE #207: Show the reports popup handle regardless of whether there are reports
Right now the behaviour is to hide the "view past reports" button if there are none. This is misleading users into thinking that there is no way of saving the past performance when they initially access the website (which is when their first impression and strongest beliefs form).

Improvement: change the wording of the button from "view past reports" to "performance reports", and display that button regardless of the existence of reports. If there are none, show an empty state with a brief prompt encouraging to leave performance reports

aabounegm:
Resolves #207


illright:
Just figured I'd make it look a little nicer
image

PR #224, opened 2020-04-25
Copy dates and enable Telegram by default

illright:


src/routes/projects/new.svelte
```diff
@@ -229,6 +227,17 @@
     } catch (e) {
       console.error(e);
     }
+
+    prepareAfterBackend(updatedActivity);
+    for (let activity of $project.activities) {
+      if (activity._type === ActivityTypes.TEMPLATE && activity.timeframe == null) {
+        activity.timeframe = {
+          start: updatedActivity.timeframe.start,
+          end: updatedActivity.timeframe.end,
```
aabounegm, line 236: Could be simplified to 
```suggestion
          ...updatedActivity.timeframe,
```
but not a big deal

illright: Meh, let's just be explicit

PR #226, opened 2020-04-25
Scroll the duplicated activities into view automatically

### LINKED ISSUE #206: Clicking Copy Activity on the View Project page is misleading
The copied activities are appended to the end. This is especially a problem when copying activities from the start or the middle, where the end of the list is invisible. Gives a feeling like nothing is happening on the page. Perhaps an animation or a scroll-down to the new activity can be good enough a hint of what's happening on the page.

### LINKED ISSUE #211: Allow editing performance reports on volunteers
For that we'd have to leave the "create report" button in place, which also helps with avoiding the empty panel issue. Sweet!

illright:
Closes #206 
Upd: and accidentally also closes #211

### DIFF
--- a/src/components/projects/view/report-performance-modal.svelte
+++ b/src/components/projects/view/report-performance-modal.svelte
@@ -11,15 +11,20 @@
 
   export let activity;
   export let application;
+  export let report;
   export let isOpen = false;
 
-  let value = {
+  $: value = report || {
     rating: null,
     content: '',
   };
 
   function submit() {
-    dispatch('submit', { value, application, activity });
+    value = {
+      rating: value.rating,
+      content: value.content,
+    };
+    dispatch('submit', { value, application, activity, report });
     value = {
       rating: null,
       content: '',
@@ -32,7 +37,7 @@
 <Modal bind:isOpen>
   {#if application != null}
     <Dialog
-      title="Create a report on {application.applicant.full_name}"
+      title="{report == null ? 'Create a' : 'Edit the'} report on {application.applicant.full_name}"
       classname="report-performance"
       closeCallback={() => isOpen = false}
     >
@@ -61,7 +66,11 @@
           </Labeled>
           <Labeled icon label="Actual worktime">
             <svg class="icon" slot="icon" src="images/icons/clock.svg" />
-            {application.actual_hours} hour{s(application.actual_hours)}
+            {#if activity.fixed_reward}
+              as needed
+            {:else}
+              {application.actual_hours} hour{s(application.actual_hours)}
+            {/if}
           </Labeled>
         </div>
         <label for="rating">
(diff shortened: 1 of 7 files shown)


src/components/projects/view/report-performance-modal.svelte
```diff
@@ -11,15 +11,20 @@
...
 
-  let value = {
+  $: value = report || {
     rating: null,
     content: '',
   };
 
   function submit() {
-    dispatch('submit', { value, application, activity });
+    value = {
```
aabounegm, line 23: Assigning to a reactive variable, and the assignment also happens to just be `value = {...value}`. I don't understand this.

src/routes/projects/[id([0-9]+)]/index.svelte
```diff
@@ -236,15 +236,20 @@
...
       reportPerformanceModal.application = detail.application;
+      reportPerformanceModal.report = detail.report;
       reportPerformanceModal.open = true;
     },
     async submitReport({ detail }) {
       try {
-        const { value, activity, application } = detail;
-        application.reports.push(await api.json(api.post(
+        const { value, activity, application, report } = detail;
+        const apiCall = report == null ? api.post : api.patch;
```
Comment (aabounegm) [inline src/routes/projects/[id([0-9]+)]/index.svelte:251]: Could just add `PUT` method in backend, but it doesn't matter much)

Comment (illright) [reply src/routes/projects/[id([0-9]+)]/index.svelte:251]: Let's be consistent

src/components/projects/view/report-performance-modal.svelte
illright: This is to unpack the `rating` and `content` fields of the `report` object from the backend, leaving all other fields behind. Maybe not the most elegant, but seems to work.

illright:
> If you copy an activity and then delete the original one, the copies also get deleted. This is inconsistent with what happens after clicking "edit" on the original activity and then discarding it anyway, where the copy persists in edit mode. Is this intended behavior?

Cannot reproduce. Perhaps you could give a more detailed sequence of actions?

> I cannot discard the first activity in edit mode (not yet created) even if there are other activities in the template.

Yeah, that is sort of intended, although strange and ugly. I'll just blindly blame it all on the architecture.

> Also, I was somehow able to produce a bug where an activity is repeated twice

This is indeed a bug, and a very nasty one. I'll try to reproduce it now, but if you could give me a hand here, this would also be helpful. Note that this bug usually comes with some error in the console.

src/components/projects/view/report-performance-modal.svelte
aabounegm: I don't see why it cannot be inlined or even given another name. My guess as to why it works is that reactivity doesn't kick in immediately, but gets scheduled in the event loop which doesn't interrupt this function, so it dispatches with the modified value (which will get reset to the reactive value after this function). Also, the same comment applies to the assignment after dispatching

PR #233, opened 2020-05-01
Feature/reclaim innopoints

### LINKED ISSUE #231: Create a form to reclaim old innopoints
This could be in the form of a button in the profile that opens a modal with two fields: username and password. After filling them out, the processing takes place, after which the modal should close (but the button to open it again should remain) and the transferred innopoints should be immediately visible in the profile.

illright:
Closes #231 
Keep in mind that the backend is not yet ready.


src/routes/profile.svelte
```diff
@@ -102,15 +103,41 @@
...
+      try {
+        const resp = await api.post('/reclaim-innopoints', { data: {
+          email: detail.email,
+          password: detail.password,
+        }});
+
+        if (resp.status === 403) {
+          reclaimInnopointsModal.error = 'Incorrect username or password.';
+        } else if (!resp.ok) {
+          console.log(resp.status);
```
aabounegm, line 122: Use `console.error`, preferably with `await resp.text()` or `(await resp.json()).message`

src/routes/profile.svelte
```diff
@@ -102,15 +103,41 @@
...
+    error: null,
+    show() {
+      reclaimInnopointsModal.open = true;
+    },
+    async reclaimInnopoints({ detail }) {
+      try {
+        const resp = await api.post('/reclaim-innopoints', { data: {
+          email: detail.email,
+          password: detail.password,
+        }});
```
aabounegm, line 117:
```suggestion
        const resp = await api.post('/reclaim-innopoints', { data: detail });
```

PR #235, opened 2020-05-02
CSRF protection

### LINKED ISSUE #234: Set up CSRF protection
The server will generate a CSRF token for every session and return it as a field of `/account`. That token is to be passed back to the server for every modifying request through the `X-CSRF-Token` header.

### LINKED ISSUE #237: Color radiobuttons in the Create Product preview card are broken
A 5-seconds style fix, probably

illright:
Closes #234 
Also a nice reason to update all API requests to our latest patterns
Also closes #237


src/routes/products/[id([0-9]+)]/edit.svelte
```diff
@@ -66,13 +66,23 @@
   let errorMessage = null;
   let warningDialogOpen = false;
 
-  async function addColor(color) {
-    let resp = await api.post('/colors', { data: { value: color } });
-    if (resp.status == 200) {
+  let colorDebounce = false;
+  async function addColor({ detail: color }) {
+    if (colorDebounce || colors.include(color)) {
```
Comment (aabounegm) [inline src/routes/products/[id([0-9]+)]/edit.svelte:71]: Your good sense of grammar caused a typo here
```suggestion
    if (colorDebounce || colors.includes(color)) {
```

src/routes/products/[id([0-9]+)]/edit.svelte
```diff
@@ -66,13 +66,23 @@
...
+      await api.json(api.post(
+        '/colors',
+        { data: { value: color }, csrfToken: account.csrf_token },
+      ));
       colors.push({ value: color });
       colors = colors;
-    } else {
-      console.error(`Color creation failed (${resp.status})`);
+      colorDebounce = true;
+      setTimeout(() => colorDebounce = false, 200);
```
Comment (aabounegm) [inline src/routes/products/[id([0-9]+)]/edit.svelte:83]: I couldn't reproduce Max's browser's behavior (of firing continuously), but it seems to me that this code implements throttle and not debounce

src/routes/profile.svelte
```diff
@@ -32,6 +32,7 @@
     ]));
 
     data.timelineFetchedUntil = timelineFetchedUntil;
+    data.account.csrf_token = session.account.csrf_token;
```
aabounegm, line 35: Assigning to `data.account.csrf_token` right before checking if `data.account == null`?

src/routes/profile.svelte
```diff
@@ -131,22 +138,17 @@
...
-      }
-    });
+  async function changeUsername({ detail }) {
+    try {
+      await api.json(api.patch(`/accounts/${account.email}/telegram`, {
+        data: { telegram_username: detail },
+        csrfToken: account.csrf_token,
+      }));
+      account.telegram_username = detail;
+      $session.account = account;
```
aabounegm, line 148: You are overriding the logged in user with the user whose profile is currently being viewed. I logged in as an admin, viewed a student's profile, changed their telegram, and all of a sudden I became the student!

src/routes/projects/[id([0-9]+)]/index.svelte
```diff
@@ -65,15 +67,24 @@
...
 
         if (telegram && telegram != account.telegram) {
-          api.patch('/account/telegram', { data: { telegram_username: telegram } });
+          try {
+            await api.json(api.patch(
+              '/account/telegram',
+              { data: { telegram_username: telegram }, csrfToken: account.csrf_token },
+            ));
+            account.telegram = telegram;
+            $session.account = account;
```
Comment (aabounegm) [inline src/routes/projects/[id([0-9]+)]/index.svelte:84]: Comment from profile doesn't apply here, but perhaps it is better to use a store method like `.set` or `.update`?

src/routes/products/[id([0-9]+)]/edit.svelte
Comment (illright) [reply src/routes/products/[id([0-9]+)]/edit.svelte:71]: :)

src/routes/products/[id([0-9]+)]/edit.svelte
Comment (illright) [reply src/routes/products/[id([0-9]+)]/edit.svelte:83]: Dammit, not again. Yeah, I couldn't reproduce the thing either, it's probably just the Mac's system color picker.

src/routes/projects/[id([0-9]+)]/index.svelte
Comment (illright) [reply src/routes/projects/[id([0-9]+)]/index.svelte:84]: Why? It would be more verbose. I understand that it ensures no subscriptions, but then again, perhaps the Svelte compiler is smart enough to recognize this. So far I see no reason for it.

PR #239, opened 2020-05-04
Prohibit whitespace-only names

### LINKED ISSUE #236: Prohibit whitespace-only names
Those are not rendered to HTML, so it messes with the markup at places. Generally not a good idea to allow such things.

illright:
Closes #236 
Will also rename the Moderation internal activity to `[[Moderation]]`, because I figured that in some cases people would wanna create the Moderation activity for themselves, so to avoid collisions I added symbols.

### DIFF
--- a/src/components/projects/new/edit-activity.svelte
+++ b/src/components/projects/new/edit-activity.svelte
@@ -63,7 +63,7 @@
   }
 
   function validateAndSave() {
-    if (!activity.name) {
+    if (!activity.name || /^\s*$/.test(name)) {
       errors.nameNotSpecified = true;
     }
 
--- a/src/components/projects/new/edit-activity.svelte
+++ b/src/components/projects/new/edit-activity.svelte
@@ -67,7 +67,7 @@
       errors.nameNotSpecified = true;
     }
 
-    if (activity.name === 'Moderation') {
+    if (activity.name === '[[Moderation]]') {
       errors.nameReserved = true;
     }
 
--- a/src/components/projects/new/edit-activity.svelte
+++ b/src/components/projects/new/edit-activity.svelte
@@ -63,7 +63,7 @@
   }
 
   function validateAndSave() {
-    if (!activity.name || /^\s*$/.test(name)) {
+    if (!activity.name || /^\s*$/.test(activity.name)) {
       errors.nameNotSpecified = true;
     }
 
@@ -118,7 +118,7 @@
       title="Activity name"
       required
       error={
-        (activity.name === '' && "The name must not be empty.")
+        (/^\s*$/.test(activity.name) && "The name must not be empty.")
         || (errors.nameNotSpecified && "The name must be specified.")
         || (errors.nameReserved && "This name is reserved for internal use.")
         || (!ensureNameUniqueness() && "The name should be unique.")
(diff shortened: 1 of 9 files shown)


src/components/projects/new/edit-activity.svelte
```diff
@@ -63,11 +63,11 @@
   }
 
   function validateAndSave() {
-    if (!activity.name) {
+    if (!activity.name || /^\s*$/.test(activity.name)) {
```
aabounegm, line 66: Regex is slower than native methods
```suggestion
    if (!activity.name.trim()) {
```

src/components/projects/new/edit-activity.svelte
```diff
@@ -118,7 +118,7 @@
       title="Activity name"
       required
       error={
-        (activity.name === '' && "The name must not be empty.")
+        (/^\s*$/.test(activity.name) && "The name must not be empty.")
```
aabounegm, line 121: Same here
```suggestion
        (!activity.name.trim() && "The name must not be empty.")
```

src/containers/projects/new/step-1.svelte
```diff
@@ -25,7 +25,7 @@
     required
     error={
          (duplicateName && "The name must be unique.")
-      || ($project.name === '' && "The name must not be empty.")
+      || (/^\s*$/.test($project.name) && "The name must not be empty.")
```
aabounegm, line 28:
```suggestion
      || (!$project.name.trim() && "The name must not be empty.")
```

src/routes/products/[id([0-9]+)]/edit.svelte
```diff
@@ -190,14 +190,14 @@
   }
 
   async function saveChanges() {
-    if (!product.name) {
+    if (!product.name || /^\s*$/.test(product.name)) {
```
Comment (aabounegm) [inline src/routes/products/[id([0-9]+)]/edit.svelte:193]:
```suggestion
    if (!product.name.trim()) {
```

src/routes/products/new.svelte
```diff
@@ -135,14 +135,14 @@
   async function createProduct() {
     errorMessage = null;
     prefetch('/store');
-    if (!product.name) {
+    if (!product.name || /^\s*$/.test(product.name)) {
```
aabounegm, line 138:
```suggestion
    if (!product.name.trim()) {
```

src/routes/projects/[id([0-9]+)]/edit.svelte
```diff
@@ -143,7 +143,7 @@
         required
         error={
              (duplicateName && "The name must be unique.")
-          || (project.name === '' && "The name must not be empty.")
+          || (/^\s*$/.test(project.name) && "The name must not be empty.")
```
Comment (aabounegm) [inline src/routes/projects/[id([0-9]+)]/edit.svelte:146]:
```suggestion
          || (!project.name.trim() && "The name must not be empty.")
```

src/routes/products/[id([0-9]+)]/edit.svelte
Comment (aabounegm) [reply src/routes/products/[id([0-9]+)]/edit.svelte:193]: Also, if `product.name` can be `null`, the RegEx test can deal with it. No need to check it first

Comment (illright) [reply src/routes/products/[id([0-9]+)]/edit.svelte:193]: But it will say `false`, while I want it to say `true` in that "true, this is a bad name"

PR #256, opened 2020-05-29
Attractions migration

### LINKED ISSUE #103: Add the "Activity deleted, undo" notice to the activity form
In Create Project, Step 2, when you delete an activity, it shouldn't be deleted right away, but rather scheduled for deletion with an option to undo the deletion, as shown in the markup.

**Danger point**: the scheduled deletion should be force-executed when changing steps or leaving the page.

### LINKED ISSUE #130: Indicate the varieties that went out of stock
Currently, when the user is on the View Product page and they select a variety that went out of stock, the purchase button gets disabled and a message is displayed saying that a variety is out of stock. This requires an action only to discover that the desired variety is not there.

Improvement:
- disable the respective radio buttons
- display the global "out of stock" notice when all varieties are out of stock, and disable the purchase button

### LINKED ISSUE #135: Add confirmation dialogs for dangerous actions in activity manipulation
The pages like Step 2 of Create Project and the Ongoing Moderator View for View Project have buttons like "discard changes" and "discard activity". In markup, these actions are protected by dialogs (IDs: `confirm-discard-changes`, `confirm-discard-activity` in View Project).

### LINKED ISSUE #202: Explicitly write out the optionality of certain fields
*sigh* Never thought I'd see the day.

The fields in the Create Activity form could use some clarity in what's optional

### LINKED ISSUE #218: Activity deletion is not clear
We should either animate the deletion somehow or replace the deleted card with a prompt to restore it, as suggested in #103 (the preferred method)

### LINKED ISSUE #238: The activity list in Create Project shows duplicates again
I am **so sick** of this bug. This time my course of actions was something of the following:
- create a project from a template
- edit the second (or third) activity (it was also fixed-reward) to create it from the template
- edit another activity above it
- edit that first activity again to set a different date
- copy an uncreated template activity and create a copy
- delete an uncreated template activity

### LINKED ISSUE #254: Migrate to Attractions

illright:
Closes #254 
Closes #103 
Closes #218 
Closes #238 
Closes #202 
Closes #135 
Closes #130 

Checklist:
- [x] Home
- [x] Projects
- [x] Past Projects
- [x] Store
- [x] Profile
- [x] Admin Dashboard
- [x] View Project
- [x] Review Project
- [x] Create Project
- [x] Edit Project
- [x] View Product
- [x] Create Product
- [x] Edit Product
- [x] Error pages
- [x] Ensure no FOUC
- [x] Ensure the vendor prefixes are set correctly


README.md
```diff
@@ -1,23 +1,41 @@
...
+yarn dev
 ```
 
-### Component system
+## Project structure
+
+There are three types of entities that build up this website:
+ - Pages ([`src/routes`](https://github.com/Innopoints/frontend/tree/master/src/routes))
+ - Containers ([`src/containers`](https://github.com/Innopoints/frontend/tree/master/src/containers))
+ - Components ([`src/components`](https://github.com/Innopoints/frontend/tree/master/src/components))
```
aabounegm, line 27: Just FYI, you could use relative paths

illright: Huh? Last I checked, it wasn't possible. Do you have any example repos in mind?

aabounegm: Our own https://github.com/Innopoints/innopoints/blob/master/README.md (the Nginx link)
And https://github.com/aabounegm/DMDI-Project/blob/master/README.md

src/components/common/notification.svelte
```diff
@@ -0,0 +1,30 @@
...
+      {/if}
+    {/each}
+  </span>
+  <Button round title="Mark as read" class="mt" on:click={() => dispatch('read', notification)}>
+    <svg src="images/icons/check.svg" class="icon" />
+  </Button>
+  <time datetime={notification.timestamp}>{formatTime(notification.timestamp)}</time>
+</li>
+
+<style src="../../../static/css/components/common/notification.scss"></style>
```
aabounegm, line 30: Does the `@` alias not work here?

src/routes/products/index.svelte
```diff
@@ -68,46 +68,45 @@
...
-      queryArgs.set('excluded_colors', JSON.stringify(excludedColors));
+      query.set('excluded_colors', JSON.stringify(excludedColors));
+    }
+    query.set('order_by', filtering.order.value.orderBy);
+    query.set('order', filtering.order.value.order);
+    query.set('limit', productLimit);
+    query.set('page', currentPage);
+
+    try {
+      ({ pages, data: products } = await api.json(api.get('/products', { query })));
```
aabounegm, line 97: Why is the statement wrapped in parentheses?

illright:
> There seems to be an inconsistency in including the styles in their components. Sometimes they are written directly in the component file, and sometimes in a separate SCSS file and linked in the component.

Only partly. I kept the styles embedded only for the components that are intended to be transferred to Attractions for easier handling. All the other components have styles separate.

> I also see a lot of :global() used. Is it really necessary?

Yes. It's okay though, because it's kept under control by the topmost selector, i.e. it's always something like this:
```css
.selector {
  .inner1 {}
  :global(.inner2) {}
}
```
and this will then be transformed into:
```css
.selector.svelte-xxx .inner1.xxx {}
.selector.svelte-xxx .inner2 {}
```
In that way the styles are still kept isolated to the component, but it allows to style children components.

src/components/common/empty-state.svelte
```diff
@@ -0,0 +1,15 @@
...
+  export let small = false;
+</script>
+
+<div class="empty" class:small>
+  <div class="icon">
+    <slot />
+  </div>
+  {#if text != null}
+    <p class="title">{text}</p>
+  {/if}
```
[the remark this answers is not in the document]
illright: For this component I don't really care what looks better, I just made it in the way that retains full generality (i.e. allows me not to dig up styles). This component will not be exposed so it's okay.

src/components/common/notification.svelte
illright: It probably does, but it's pointing to /src/ anyway, so it wouldn't be of much use. Of course, we could devise another alias, but honestly I'm thinking of removing that aliasing plugin. I just don't like it. And now that we got a lot of components from Attractions, having the aliasing for, say, `/ui/` is not that critical.

src/routes/products/index.svelte
illright: For destructuring. Braces at the beginning of the line are treated as a start of a code block, not an object.

src/components/common/empty-state.svelte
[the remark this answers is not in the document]
aabounegm: What do you mean by "This component will not be exposed"? It's not like we are building a UI library, this is Innopoints, nothing is "exposed"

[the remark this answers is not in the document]
illright: Yeah, that's why I'm opting for more convenience, perhaps at the cost of common sense

src/components/common/notification.svelte
aabounegm: Do you not like the `@` alias or the idea of aliasing in general? I think it is helpful in that we don't need to write a lot of `../../../`, and especially when moving a file to a different folder. If your problem is with the `@` symbol, it can be changed to any other symbol/word

illright: The idea of aliasing in general. As I said, I was tearing hairs from my head when I first opened the codebase because I couldn't figure anything out, and only then I asked Max and he pointed me at the rollup config. I believe no saved keystrokes are worth the huge drawback on clarity

src/components/common/empty-state.svelte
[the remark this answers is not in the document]
aabounegm: I don't know, it doesn't feel right. And how is that more convenient? Writing the text in between the tags is always more convenient than in a prop because it is much easier to notice

src/components/common/notification.svelte
aabounegm: Even if the alias was something like `src`? I'd think it seems pretty clear

src/components/common/empty-state.svelte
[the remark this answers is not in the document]
illright: The reason is actually deeper than that. I understand it doesn't make sense to you, but let me try to explain:

There are two types of empty states used around the site: regular and small. Don't let this confuse you, the markup is actually (very unfortunately) entirely different, which makes this distinction very weird. Then, the text is pretty much always the same text node, unlike the icon, which can either be an embedded SVG or a component (reminding you that components cannot be used directly in named slots). Therefore, saving the default slot for the icon makes more sense in terms of convenience

src/components/common/notification.svelte
illright: More than anything, I'd desire a `/` alias, everything else is manageable. But I believe this is not possible.

aabounegm: Why note? There doesn't seem to be a mention of such limitation in [the docs](https://github.com/rollup/plugins/tree/master/packages/alias), and the source code doesn't suggest any problems with a `/` alias. Needs to be tested

src/components/dashboard/amend-balance.svelte
```diff
@@ -0,0 +1,62 @@
...
+  const showSnackbar = getContext(snackbarContextKey);
+
+  export let users = [];
+  export let account;
+
+  let amount = '';
+
+  async function send() {
+    try {
+      const promises = users.map(user => api.json(api.patch(`/accounts/${user.email}/balance`, {
```
aabounegm, line 17: What do you say we add a parameter to the `patch`/`get`/... functions that would automatically call `json` on the response?

src/components/dashboard/amend-balance.svelte
```diff
@@ -0,0 +1,62 @@
...
+        data: {
+          change: amount,
+        },
+        csrfToken: account.csrf_token,
+      })));
+      await Promise.all(promises);
+      showSnackbar({
```
aabounegm, line 24: Wouldn't it be better if a snackbar is shown for each email so that if some succeed and others fail, it would be easy to know what happened to each and not send duplicate points?
`Promise.allSettled` can help here, but it is relatively new. Maybe attaching `.then` and `.catch` to each promise directly would be better

src/components/dashboard/send-message.svelte
```diff
@@ -0,0 +1,58 @@
...
+    <svg src="images/icons/message-circle.svg" class="icon mr-2" />
+    Send a message
+  </header>
+  <form>
+    <TextField
+      bind:value={text}
+      multiline
+      maxlength={512}
+      cols="5"
+      id="send-message"
```
aabounegm, line 47: where is this id being used? and wouldn't it cause confusion that it is the same as the class assigned to the root element?

src/containers/dashboard/tag-editor.svelte
```diff
@@ -1,19 +1,18 @@
 <script>
-  import { stores } from '@sapper/app';
-  import Card from 'ui/card.svelte';
-  import Button from 'ui/button.svelte';
-  import TextField from 'ui/text-field.svelte';
+  import { getContext } from 'svelte';
+  import { Card, Button, TextField } from 'attractions';
+  import { snackbarContextKey } from 'attractions/src/snackbar';
```
aabounegm, line 4: Am I correct to assume that importing from `src` is one of the things you are working on in the library?

src/containers/profile/info.svelte
```diff
@@ -1,70 +1,88 @@
...
   let tgChange = false;
   let tgUsername = account.telegram_username || '';
+  $: usernameValid = telegramUsernameRegex.test(tgUsername);
+  let reclaimInnopointsModalOpen = false;
 
-  function saveUsername() {
-    if (tgUsername !== '' && tgUsername !== account.telegram_username) {
-      dispatch('username-change', tgUsername);
+  async function saveUsername() {
+    if (tgUsername === '' || tgUsername === account.telegram_username) {
```
aabounegm, line 18:
```suggestion
    if (!usernameValid || tgUsername === account.telegram_username) {
```

src/containers/profile/info.svelte
```diff
@@ -1,70 +1,88 @@
...
+      await api.json(api.patch(`/accounts/${account.email}/telegram`, {
+        data: { telegram_username: tgUsername },
+        csrfToken: account.csrf_token,
+      }));
+      account.telegram_username = tgUsername;
+      if ($session.account.email === account.email) {
+        $session.account = account;
+      }
+    } catch (e) {
+      console.error(e);
```
aabounegm, line 32: How about a snackbar so no error goes unnoticed?

src/containers/profile/timeline.svelte
```diff
@@ -1,49 +1,96 @@
 <script>
   import { createEventDispatcher } from 'svelte';
-  import Button from 'ui/button.svelte';
+  import { Button, Loading } from 'attractions';
+  import EmptyState from '../../components/common/empty-state.svelte';
```
aabounegm, line 4: At least be consistent until the alias is replaced with something else so that it will be easier to use _find and replace all_
```suggestion
  import EmptyState from '@/components/common/empty-state.svelte';
```

src/containers/profile/timeline.svelte
```diff
@@ -1,49 +1,96 @@
...
+  };
+
+  const dispatch = createEventDispatcher();
 </script>
 
 <div class="timeline">
-  {#if timelinePromises.length}
+  {#if timelinePromises}
     {#each timelinePromises as promise}
       {#await promise.then(value => value.data)}
```
aabounegm, line 62: What is the use of the `.then(value => value.data)` part if the data is not being used?

src/components/dashboard/amend-balance.svelte
illright: Nah, I don't think this would help readability. There are no kwargs in JS

src/components/dashboard/amend-balance.svelte
illright: It would get really messy. Good point, but z
I think it's infeasible in practice. Besides, it shouldn't occur often that one fails and the other succeeds

src/components/dashboard/send-message.svelte
illright: It was just screaming at me for having duplicate IDs `null` for some reason

src/containers/dashboard/tag-editor.svelte
illright: Not yet, but this is something to overcome with Rollup

src/containers/profile/timeline.svelte
illright: It is used, it's building the timeline, doesn't it?

src/containers/profile/notifications.svelte
```diff
@@ -80,35 +84,56 @@
         { data: notificationSettings, csrfToken: account.csrf_token },
       ));
       changes = false;
+      const undoPrompt = showSnackbar({
+        props: {
+          text: 'Settings updated successfully!',
+          action: {
+            text: 'undo',
+            async callback() {
+              notificationSettings = Object.assign({}, initialSettings);
```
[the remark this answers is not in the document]
illright: We mutate it in the settings

package.json
```diff
@@ -16,7 +16,7 @@
     }
   },
   "lint-staged": {
-    "*.js": [
+    "[^.]*.js": [
```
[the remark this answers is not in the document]
illright: It was giving a warning about running on an excluded file, `.eslintrc.js` (I think). This avoids linting that file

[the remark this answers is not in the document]
illright: As for the dot, not when it's inside the character class, I believe. Anyway, it gets the job done

src/containers/profile/notifications.svelte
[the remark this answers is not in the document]
aabounegm: What gets mutated where?
I can't see in the code why `notificationSettings = initialSettings` is not enough.

package.json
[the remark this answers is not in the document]
aabounegm: perhaps it should not be excluded. It is a js file after all, and it should be subject to linting

src/components/dashboard/amend-balance.svelte
aabounegm: True, but many use object destructuring as an alternative for that. We already do that, with the object that has `data` and `csrfToken` as properties. Why can't there be a 3rd property `json: true`?

src/components/dashboard/amend-balance.svelte
aabounegm: I wouldn't say it is infeasible (it is very easy to implement). Perhaps it can get messy, especially when all succeed or all fail, which are the most common cases (I agree). We can check for that and collect successes/failures in one snackbar.
Alright, it's not worth it.

src/components/dashboard/send-message.svelte
illright: It seems like an SSR issue. This only happens in Sapper, and gets immediately fixed up on hydration. It's still a nasty warning in the console, so I'd like to avoid it if possible, though I must agree that this is simply treating the symptoms

illright: And it is only happening with Attractions, so I'll investigate that more. Probably a bug on our side

aabounegm: That is because of the default strigification behavior of Javascript, and the prop's [default value is `null`](https://github.com/illright/attractions/blob/develop/src/text-field/text-field.svelte#L16). This should be fixed from the component itself, and I think `id` should be part of `$$restProps`, with the label surrounding the input.
If you were thinking about empty string as default, [it is not allowed](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute)

src/containers/dashboard/tag-editor.svelte
illright: Which I could use your assistance with. I've pushed all my local changes to Attractions as a bunch of little PRs, and after merging them, maybe you could set up Rollup for it

package.json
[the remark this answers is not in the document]
illright: It is excluded by default by ESLint, it doesn't lint its own config, and the warning says that our pattern includes files that are excluded by default

src/components/dashboard/amend-balance.svelte
illright: Yeah, but this object isn't always passed, and when it is, it doesn't make stuff pretty, so I'd say that a wrapper function is cleaner. Also leads to nicer separation of concerns

src/components/dashboard/send-message.svelte
illright: No, Svelte specifically doesn't set attributes if their values are nullish

src/containers/profile/timeline.svelte
aabounegm: right. I didn't notice the `{:then timelineFragment}` part because the syntax in my mind was just `{#await promise as value}`

src/components/dashboard/amend-balance.svelte
aabounegm: I just don't like the sight of a function that never gets called without another particular function (or a few), and not having them composed to form one function that does both. Perhaps that's the effect of Haskell on me

src/components/common/notification.svelte
illright: Okay, I tried everything I could, apparently, it's that difficult to have Rollup import from project root. I couldn't get it to work, you're free to give it a shot. Note that `<style src="...">` might be resolved without Rollup, simply with Node, but that's not certain. idk how Rollup takes over file resolution.

src/containers/profile/notifications.svelte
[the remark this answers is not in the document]
illright: The `notificationSettings` gets mutated on line 73 when applying changes. If we restore the initial settings like that, we will only be able to do it once, and from then on, the `initialSettings` object will have the same changes

src/components/dashboard/amend-balance.svelte
illright: I do not share your concern, sorry)

src/containers/profile/info.svelte
illright: That thing actually serves a different purpose. If the username isn't valid, this function won't be even called. Amended

src/containers/profile/info.svelte
illright: Good call

src/components/projects/view/application-tile.svelte
```diff
@@ -1,107 +1,149 @@
...
-  export let handlePanelOpen;
-  let report = null;
 
+  let report = null;
   onMount(() => {
     report = api.get(
       `/projects/${activity.project}/activities/${activity.id}`
       + `/applications/${application.id}/report_info`,
-    ).then(resp => resp.json());
+    ).then(resp => resp.json()).catch(console.error);
```
aabounegm, line 23: Note that adding the `.catch` will make the `report` promise resolve, causing `reportData` in the template to be `undefined` in case the API call fails (because it is what `console.error` returns)

aabounegm: Also, `onMount` can take an `async` function, but that doesn't mean other code won't run until it resolves.

src/components/projects/view/application-tile.svelte
```diff
@@ -1,107 +1,149 @@
...
   });
 
+  const viewReportsDialog = {
+    open: false,
+    reports: null,
+    async show({ detail }) {
+      viewReportsDialog.reports = (await report).reports;
```
aabounegm, line 30: Now this is a cool example of when the Null Object design pattern can be useful 😉, given the previous comment

src/components/projects/view/finalizing-activity-card.svelte
```diff
@@ -0,0 +1,124 @@
...
+      hourChanges.delete(application);
+    } else {
+      hourChanges.set(application, hours);
+    }
+    hourChanges = hourChanges;
+  }
+
+  async function saveHours() {
+    const requests = [];
+    for (let [application, hours] of hourChanges) {
```
aabounegm, line 31:
```suggestion
    for (const [application, hours] of hourChanges) {
```

src/components/projects/view/tag-selector.svelte
```diff
@@ -1,21 +1,20 @@
...
 
   // eslint-disable-next-line require-yield
   async function *getOptions(query) {
+    let output = tags;
     if (query) {
-      return tags.filter(tag => tag.name.toLowerCase().includes(query.toLowerCase()));
+      output = output.filter(tag => tag.name.toLowerCase().includes(query.toLowerCase()));
     }
-    return tags;
+    return output.filter(tag => !value.includes(tag.id));
```
aabounegm, line 17: Didn't you mention in some other PR that generator functions have to always `yield` instead of `return`?

src/containers/profile/timeline.svelte
```diff
@@ -1,49 +1,80 @@
...
+  import * as api from '@/utils/api.js';
 
   export let timelinePromises;
-
-  const dispatch = createEventDispatcher();
+  export let competences;
 
   function requestMore() {
     dispatch('more-timeline');
   }
```
aabounegm, line 15: You usually inline such one-line functions. It is only used once

src/containers/profile/timeline.svelte
```diff
@@ -1,49 +1,80 @@
...
+    },
+  };
+
+  const showSnackbar = getContext(snackbarContextKey);
+  const dispatch = createEventDispatcher();
 </script>
 
 <div class="timeline">
-  {#if timelinePromises.length}
+  {#if timelinePromises}
```
aabounegm, line 45: An empty array is truthy in Javascript. Are you sure you want to remove the `.length`?

src/containers/profile/timeline.svelte
```diff
@@ -1,49 +1,80 @@
...
         </div>
       {:then timelineFragment}
         {#each timelineFragment as entry}
-          <TimelineEntry {...entry} on:leave-feedback />
+          <TimelineEntry {...entry} on:leave-feedback={leaveFeedbackModal.show} />
         {/each}
       {/await}
     {/each}
 
     {#await Promise.all(timelinePromises).then(values => values[values.length - 1]) then lastValue}
```
aabounegm, line 59: In case accepting the empty array wasn't a mistake, this here will cause an error (`values[-1] == undefined`)

src/containers/projects/new/step-2.svelte
```diff
@@ -1,123 +1,177 @@
...
+  }
+
+  function propagateDates({ detail }) {
+    for (let activity of $project.activities) {
+      if (activity.internal) {
+        continue;
+      }
+
+      if (activity.timeframe == null
+          || activity.timeframe.start == null && activity.timeframe.end == null) {
```
aabounegm, line 98: Use guard clauses. Also, this will not `patch` the date if only one of the `start` or `end` is null.

src/containers/projects/new/step-2.svelte
```diff
@@ -1,123 +1,177 @@
...
+    }
+  }
+
+  function propagateQuestions({ detail }) {
+    for (let activity of $project.activities) {
+      if (activity.internal) {
+        continue;
+      }
+
+      if (arraysEqual(activity.feedback_questions, defaultFeedbackQuestions)) {
```
aabounegm, line 118:
```suggestion
      if (!arraysEqual(activity.feedback_questions, defaultFeedbackQuestions)) continue;
```

src/containers/projects/view/project-header.svelte
```diff
@@ -1,32 +1,74 @@
...
+      $project.review_status = ReviewStatuses.PENDING;
+    } catch (e) {
+      showSnackbar({ props: { text: 'Something went wrong, try reloading the page' } });
+      console.error(e);
+    }
+  }
+
+  async function deleteProject() {
+    try {
+      prefetch('/projects');
```
aabounegm, line 62: If you prefetch before you delete the project, wouldn't that cause the page to render with that project there?

illright: Yeah, now that I think about it, `prefetch` in most places is not suitable. Will remove.

src/components/projects/view/staff-card.svelte
```diff
@@ -1,25 +1,26 @@
...
   export let moderator;
-  export let projectStage;
-  export let application;
-  export let isCreator;
-  export let review = false;
+  export let moderation;
+  export let project;
 
+  const review = getContext('review-mode');
+  const isCreator = moderator.email === $project.creator.email;
```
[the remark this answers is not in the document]
illright: `$project` is an auto-subscription to the `project` store, check out https://svelte.dev/tutorial/auto-subscriptions

src/components/projects/new/project-image-picker.svelte
```diff
@@ -1,49 +1,108 @@
...
+    uploading = false;
+    imageResizerOpen = false;
+  }
+
+  function dismissResize() {
+    inputValue = null;
+    if (uploading) {
+      uploadCancelled = true;
+      uploading = false;
+      uploadController && uploadController.abort();
```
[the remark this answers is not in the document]
illright: It's not that big of a deal, really. The reason for this abort is to allow the user with really bad internet to abort a large image upload before they waste all their bandwidth, and as for the backend, no one cares what will be there

src/constants/projects/project-templates.js
```diff
@@ -1,10 +1,10 @@
 import getBlankActivity from '@/constants/projects/blank-activity.js';
-import ActivityTypes from '@/constants/projects/activity-internal-types.js';
 
 
 export function getBlankProject() {
   return {
     name: null,
+    image_id: null,
```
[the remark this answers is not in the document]
illright: It's an object that will be sent to the backend, backend's naming conventions take precedence here.

src/components/projects/new/activity-questions.svelte
```diff
@@ -1,49 +1,58 @@
...
+    <Dropdown top>
+      <Label class="mb">Feedback questions</Label>
+      {#each value as question, i}
+        <div class="row">
+          <TextField value={question} maxlength={1024} on:change={(e) => changeValue(i, e)} />
+          <Button
+            danger
+            round
+            title="Delete question"
+            on:click={(e) => { e.stopPropagation(); removeValue(i); }}
```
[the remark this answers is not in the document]
illright: This only works for DOM events, it won't let us attach modifiers (except for `once`) to custom events.

src/components/common/stars.svelte
```diff
@@ -0,0 +1,18 @@
+<script>
+  import range from '@/utils/range.js';
+
+  export let value;
+  export let max = 5;
+</script>
+
+<div class="stars">
+  {#each [...range(max)] as i}
+    {#if i < value}
```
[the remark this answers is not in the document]
illright: Remember that SVGs are actually substituted, I wouldn't play with it much. It might work, but it's best to do this if we will ever move to component-based icons

src/utils/project-manipulation.js
```diff
@@ -8,202 +9,169 @@ export function copyActivity(activity) {
...
-    moderators: project.moderators,
-  };
-  if (!edit) {
-    filtered.activities = project.activities.filter(act => act._type != ActivityTypes.TEMPLATE);
+/* Copy ALL the fields from the source activity object to the destination in place,
+   deep-copying complex fields. */
+export function copyActivityInPlace(source, destination) {
+  for (let field in source) {
+    destination[field] = source[field];
   }
```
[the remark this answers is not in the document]
illright: I guess it isn't

src/utils/project-manipulation.js
```diff
@@ -8,202 +9,169 @@ export function copyActivity(activity) {
...
-    }
-    insertionIndex++;
+/* Set a field on the object, creating a blank object if the destination is null. */
+function setNullSafe(object, field, value) {
+  if (object == null) {
+    return { [field]: value };
+  } else {
+    object[field] = value;
+    return object;
   }
```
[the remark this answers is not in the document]
illright: It's great that it's shorter, but in this case I think my version conveys the intent better.

src/utils/autocomplete-users.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function autocompleteValueToUser(value) {
+  return {
+    full_name: value.name,
+    email: value.details,
+  };
+}
+
+export async function* getUsersRaw(query, filterFunction = null) {
```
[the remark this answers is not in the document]
illright: Thought about both options, having `null` seems both faster and more respectful to the "nothing" concept

src/utils/autocomplete-users.js
```diff
@@ -0,0 +1,37 @@
...
+  let pages, data;
+  while (pages == null || currentPage <= pages) {
+    ({ pages, data } = await api.json(api.get('/accounts', {
+      query: new Map([['q', query], ['page', currentPage]]),
+    })));
+    if (typeof filterFunction === 'function') {
+      yield data.filter(filterFunction).map(userToAutocompleteValue);
+    } else {
+      yield data.map(userToAutocompleteValue);
+    }
```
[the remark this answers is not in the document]
illright: You really like shortening code, don't you. Okay, this one's fair

src/components/projects/new/edit-activity-card.svelte
```diff
@@ -0,0 +1,309 @@
...
+        });
+        return;
+      }
+      activity.vacant_spots -= approved.length;
+    }
+    activity.people_required = detail.value;
+  }
+
+  function validateApplicationDeadline({ detail }) {
+    if (activity.applications != null) {
```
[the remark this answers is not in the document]
illright: This one's bringing some duplication for the purpose of moving code down one indent, I don't think it's worth it

src/components/projects/new/steps-navigation.svelte
```diff
@@ -1,24 +1,30 @@
...
+    <Button class="handle" on:click={toggle}>
+      <span class="step-index"><span class="step">step</span> {step}</span>
+      <span class="step-name">{stepNames[step]}</span>
+      <svg class="ml dropdown-chevron" src="images/icons/chevron-down.svg" />
     </Button>
-  {/each}
-</Dropdown>
+    <Dropdown>
+      {#each stepNames as stepName, index}
+        <Button selected={step === index} href="/projects/new{index > 0 ? `?step=${index}` : ''}">
```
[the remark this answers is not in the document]
illright: We are, it resets if you try to bypass the step 0, but other than that, it commands the open step on the page

src/components/projects/view/application-hour-tile.svelte
```diff
@@ -1,108 +1,141 @@
...
+      application.reports = application.reports.filter(
+        report => thisReport.reporter_email !== account.email,
+      );
+      const snackbar = showSnackbar({
+        props: {
+          text: 'Report deleted',
+          action: { text: 'undo' },
+        },
+      });
+      if (await snackbar.expired) {
```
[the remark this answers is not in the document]
illright: It is, and it means that the deletion can now be carried out for sure. Before that, it doesn't delete

src/components/projects/view/application-tile.svelte
illright: Just switched it to `api.json` and removed the catch. Seems to be working fine when the server rejects that endpoint

src/components/projects/view/application-tile.svelte
illright: And this will not even be called if the promise fails, because the button won't be there

src/components/projects/view/finalizing-activity-card.svelte
illright: I've been using `let` in loops all the time, doesn't `const` have the same scope shortcomings as `var`?

src/components/projects/view/tag-selector.svelte
illright: Good call

src/containers/profile/timeline.svelte
illright: This was there from a long time ago. I think I did it this way back then to emphasize what the function is doing.

src/containers/profile/timeline.svelte
illright: Am not

src/containers/projects/new/step-2.svelte
illright: Guard clauses?
As for not patching, that's intentional

src/containers/projects/new/step-2.svelte
illright: Also not worth it. The indented block isn't that large, and inverting the condition affects understanding

src/components/projects/view/staff-card.svelte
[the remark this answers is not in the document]
aabounegm: I know about that, but I thought there had to be a line like `const { project } = stores();` or something similar. Or is it because the `project` prop is already such store?

src/components/common/stars.svelte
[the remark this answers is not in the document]
aabounegm: And what about the [`StarRating` component](https://github.com/illright/attractions/blob/develop/attractions/star-rating/star-rating.svelte)?

src/utils/autocomplete-users.js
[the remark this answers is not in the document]
aabounegm: Less code == less place for bugs (and is usually cleaner and easier to follow) ¯\_(ツ)_/¯

src/components/projects/new/edit-activity-card.svelte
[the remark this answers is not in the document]
aabounegm: So was I correct to assume that we needed to trigger reactivity in that case, and not just exit the function? (inferred it from the code)

src/components/projects/view/finalizing-activity-card.svelte
aabounegm: Surely not. `const` (like `let`) creates a new block scope with each iteration. Otherwise, it would have been like reassigning to the same variable, which can't be done to a constant.

src/components/products/new/color-picker.svelte
```diff
@@ -0,0 +1,82 @@
...
+        const payload = { value: colorValue };
+        await api.json(api.post(
+          '/colors',
+          { data: payload, csrfToken: $session.account.csrf_token },
+        ));
+        $colors.push(payload);
+        $colors = $colors;
+        colorDebounce = null;
+      } catch (e) {
+        console.error(e);
```
aabounegm, line 40: Snackbar would be better

src/utils/autocomplete-users.js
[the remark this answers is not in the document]
aabounegm: What's up with the "nothing" concept? Here you are defining a default value. If it was "nothing", it would be `undefined`, which is Javascript's default value when you pass "nothing" (don't pass anything) for that argument. A default value is a value to fallback to when "nothing" is given.

src/components/projects/view/application-tile.svelte
aabounegm: Yeah, but that was relevant to the `catch` resolving thing.

aabounegm: Yeah, but the comment was more relevant with the `catch` resolving thing

src/components/projects/new/edit-activity-card.svelte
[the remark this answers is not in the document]
illright: Not only reactivity, we have to set a value from the `detail`

src/components/common/stars.svelte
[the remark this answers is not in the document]
illright: `StarRating` is interactive, it's meant to select a rating rather than display it. This one only displays

src/components/projects/view/staff-card.svelte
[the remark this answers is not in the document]
illright: Yes, it takes it from a prop.

src/containers/errors/error-500.svelte
```diff
@@ -0,0 +1,27 @@
...
+</script>
+
+<div class="material padded">
+  {#if dev}
+    <H1>Alright, so here's what happened</H1>
+    {#if error.stack}
+      <pre>{error.stack}</pre>
+    {/if}
+  {:else}
+    <H1>Everything is fine</H1>
```
aabounegm, line 15: Reminds me of this 😂 
image
Actually, how about we add it there 😃?

src/containers/errors/error-404.svelte
```diff
@@ -0,0 +1,57 @@
...
+      alt="Go to the projects"
+    />
+  </a>
+  <a href="/products" class="destination">
+    <img
+      src="/images/404/browser.svg"
+      width={cardWidth}
+      height={cardHeight}
+      class="repr"
+      alt="Go to a product page"
```
aabounegm, line 43: "**a** product", but it takes you to `/products`?

src/components/common/stars.svelte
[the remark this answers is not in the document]
aabounegm: maybe it is worth adding a `readOnly` prop that would just display it?

src/containers/errors/error-500.svelte
illright: Oh, darn, it reads "this is fine". This is exactly what I was referring to, but if you want to, we can add a picture too :p

aabounegm: I think it would be funny)

src/components/common/stars.svelte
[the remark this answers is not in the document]
illright: Nah, those are completely different components with different markup. Perhaps what we could do is move the read-only stars to Attractions

src/routes/projects/[id([0-9]+)]/index.svelte
```diff
@@ -43,6 +42,12 @@
     account != null
     && (account.is_admin || moderatorEmails.includes(account.email))
   );
+  $: firstActivities = (
+    $project.activities.slice(0, 3)
+      .filter(activity => !activity.internal)
```
Comment (aabounegm) [inline src/routes/projects/[id([0-9]+)]/index.svelte:47]: How about `filter` before `slice`? If the first few ones happen to be internal (we have nothing in the DB that prevents this, afaik), This will limit the visible activities

src/routes/projects/past.svelte
```diff
@@ -27,10 +26,10 @@
   export let projects;
   export let account;
   export let pages;
-  let currentPage = 1;
+  const currentPage = 1;
```
aabounegm, line 29: How would the **current** page get updated if it is a `const`? The pagination component has been updated to react (bind) to `currentPage`

src/routes/projects/[id([0-9]+)]/index.svelte
Comment (illright) [reply src/routes/projects/[id([0-9]+)]/index.svelte:46]: True

src/components/products/product-card.svelte
```diff
@@ -21,7 +21,7 @@
       : '/images/create-product/placeholder.svg'
   );
 
-  $: outOfStock = varieties.reduce((sum, variety) => variety.amount + sum, 0) === 0;
+  $: outOfStock = varietiesByColor.get(selectedColor)[0].amount === 0;
```
illright, line 24: Doesn't this mean "if the first size (can be any) of the selected color is out of stock, display the notice"?
Perhaps we'd be better off keeping the reduce:
```js
varietiesByColor.get(selectedColor).reduce((sum, variety) => variety.amount + sum, 0) === 0
```

PR #264, opened 2020-07-03
Allow clearing the color selection

### LINKED ISSUE #261: Allow clearing the color selection for varieties
Currently the only way to do that is to recreate the variety

illright:
Closes #261

### DIFF
--- a/src/components/products/new/color-picker.svelte
+++ b/src/components/products/new/color-picker.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { getContext } from 'svelte';
+  import { getContext, createEventDispatcher } from 'svelte';
   import { stores } from '@sapper/app';
   import { DropdownShell, Dropdown, Button, RadioGroup, Divider } from 'attractions';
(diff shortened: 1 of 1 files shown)


src/components/products/new/color-picker.svelte
```diff
@@ -61,6 +62,11 @@
       />
     </Button>
     <Dropdown>
+      {#if value != null}
+        <Button on:click={() => { value = null; dispatch('change'); }}>
```
aabounegm, line 66: And who is listening to the event?

illright: https://github.com/Innopoints/frontend/blob/master/src/components/products/new/variety-form.svelte#L57

aabounegm: huh, because the event wasn't being dispatched previously, and I don't see it being dispatched anywhere else. Shouldn't it get dispatched on any color selection?

illright: It gets forwarded on color selection

PR #269, opened 2020-07-04
Implement the resizing dialog for several images

### LINKED ISSUE #134: Enforce aspect ratio and minimum size on uploaded images
To protect the grid from breaking, it might be useful to add an online cropping tool that will ensure a fixed aspect ratio (1:1 or 16:9, based on the situation). 
Also one could scale the low-resolution images using CSS so that the cards would not shrink.

### LINKED ISSUE #262: Create an image resizing dialog that is capable of processing several images
It could have image miniatures at the bottom (perhaps even with uploading indicators to be able to show the dialog right away), with checkmarks marking the images that were cropped successfully

illright:
Phew, that was messy. Test this one carefully.

Closes #262
Closes #134

### DIFF
--- a/src/components/common/multi-image-resizer.svelte
+++ b/src/components/common/multi-image-resizer.svelte
@@ -44,7 +44,7 @@
     }
   }
 
-  function dismissImage(image = null) {
+  function dismissImage(image) {
     image.reject();
     images = images.filter(thatImage => thatImage !== image);
     repositionCursor();
--- a/src/components/common/multi-image-resizer.svelte
+++ b/src/components/common/multi-image-resizer.svelte
@@ -1,0 +1,168 @@
+<script>
+  import { getContext } from 'svelte';
+  import { stores } from '@sapper/app';
+  import { Button, Dialog, Modal, Divider, Loading } from 'attractions';
+  import { snackbarContextKey } from 'attractions/snackbar';
+  import Cropper from 'svelte-easy-crop';
+  import * as api from 'src/utils/api.js';
+
+  const { session } = stores();
+
+  export let aspectRatio = 1;
+  export let open = false;
+
+  let currentImage = 0;
+  let images = [];
+
+  const crop = { x: 0, y: 0 };
+
+  function saveBounds({ detail }) {
+    images[currentImage].pixels = detail.pixels;
+  }
+
+  function resetCropArea() {
+    crop.x = 0;
+    crop.y = 0;
+  }
+
+  async function repositionCursor() {
+    for (let idx = 0; idx < images.length; ++idx) {
+      if (images[idx].uploadPromise == null) {
+        currentImage = idx;
+        resetCropArea();
+        return;
+      }
+    }
+
+    try {
+      currentImage = null;
+      await Promise.all(images.map(image => image.uploadPromise));
+    } finally {
+      open = false;
+      images = [];
+      currentImage = 0;
+    }
+  }
+
+  function dismissImage(image = null) {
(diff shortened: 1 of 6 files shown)


src/components/common/multi-image-resizer.svelte
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  function resetCropArea() {
+    crop.x = 0;
+    crop.y = 0;
+  }
+
+  async function repositionCursor() {
+    for (let idx = 0; idx < images.length; ++idx) {
+      if (images[idx].uploadPromise == null) {
```
aabounegm, line 30:
```suggestion
    if (images.some(image => image.uploadPromise == null)) {
```
Or even a `for .. of` loop

src/components/common/multi-image-resizer.svelte
```diff
@@ -0,0 +1,168 @@
...
+      currentImage = null;
+      await Promise.all(images.map(image => image.uploadPromise));
+    } finally {
+      open = false;
+      images = [];
+      currentImage = 0;
+    }
+  }
+
+  function dismissImage(image = null) {
```
aabounegm, line 47: Why does it default to `null`? If `null` is an acceptable value, it needs to be checked for in the body

src/components/common/multi-image-resizer.svelte
```diff
@@ -0,0 +1,168 @@
...
+    images = images;
+
+    return new Promise((resolve, reject) => {
+      imageObject.resolve = resolve;
+      imageObject.reject = reject;
+    });
+  }
+
+  function alertOnClose({ detail }) {
+    if (!detail.value) {
```
aabounegm, line 96:
```suggestion
    if (detail.value) return;
```

src/components/common/multi-image-resizer.svelte
illright: We need the index

src/components/common/multi-image-resizer.svelte
illright: Oh, I just forgot to remove it

src/components/common/multi-image-resizer.svelte
illright: I'm always hesitant to accept suggestions like this, but this time especially. `detail.value` is the `open` value of the modal, and this way of writing it makes it clear that we're only acting upon the _close_ event

illright:
> Is there a way not to close the modal when clicking outside?

Not yet, this is something we should add to Attractions

> Also, shouldn't the "delete" button on images inside the dropzone (after uploading) actually delete it from the server?

Ideally, it should, but I couldn't be bothered, this component was hard enough

src/components/common/multi-image-resizer.svelte
aabounegm: `function alertOnClose({ detail: { value: open } })`, followed by `if(!open) return;` in the body should make it clear enough

illright: Yeah, but is all of this mess worth that tiny indent reduction? I don't think so
I know about the cognitive complexity, but I don't take it seriously in code like this (as opposed to complex view logic on the backend where it all does add up)

aabounegm: And even if not for the guard clause, the rename makes it clearer

illright:
Actually, before this, it's best if we merge the PR in Attractions, release and add a commit here that bumps dependencies

PR #280, opened 2020-07-18
Bump the dependencies

illright:
Also remedies a security vulnerability


illright:
There's gonna be a warning in the console when building, it can be safely ignored

aabounegm:
Isn't the main purpose of this PR to update the Svelte version? Why did it not change in package.json?

illright:
It was a minor upgrade, no change in package.json necessary

PR #293, opened 2020-09-22
Replace the fonts for the guideline ones

illright:
where men cried


aabounegm (approved):
image

(Also, didn't run it but code looks good)

PR #294, opened 2020-09-29
Create a maintenance page

illright:
Care must be taken when actually deploying this page because it contains an `` reference to a local file


aabounegm:
How about putting both files in one folder such that it can be easily used with nginx's `root`, for example?

illright:
Well, this maintenance page is kind of its own entity and I didn't wanna mess with the structure much, so we'll handle the file moving on deployment

aabounegm:
That's the point. since it is its own entity, move it out of `src` and into `maintenance` (or something similar) since it doesn't depend on anything in the Sapper project, and to keep it all self-contained (not one HTML page somewhere, referring to an image in a very different folder)

aabounegm (approved):

PR #295, opened 2020-10-29
Bump dependencies

illright:
Finally everything seems fine with the CSS after the update. I'm planning to work on the statistics to see if any of it needs fixing, figured I'd do this first


rollup.config.js
```diff
@@ -48,14 +47,19 @@ export default {
       }),
       eslint(),
       svelte({
-        extensions: ['.html', '.svelte', '.svg'],
+        extensions: ['.html', '.svelte'],
         preprocess: preprocessChain,
         dev,
         hydratable: true,
       }),
+      // TODO: figure out the new image caching mechanism
```
aabounegm, line 55: What exactly about it is to be figured out?

aabounegm (commented):
Did you test it or should I? Especially the Docker images

rollup.config.js
illright: Well, do we implement it or not. I think we do, I'll add some more commits in a bit

illright: Nah, screw it

aabounegm (approved):

PR #301, opened 2020-12-04
Bug report: cannot set zero hours for volunteers in the finalizing stage

illright:
Check out the last commit, please. It's not directly related to the issue but something was odd there and I changed it without testing, so I need a sanity check

### DIFF
--- a/src/components/projects/view/application-hour-tile.svelte
+++ b/src/components/projects/view/application-hour-tile.svelte
@@ -32,7 +32,7 @@
     try {
       const thisReport = report;
       application.reports = application.reports.filter(
-        report => thisReport.reporter_email !== account.email,
+        report => thisReport.reporter_email !== report.reporter_email,
       );
       const snackbar = showSnackbar({
         props: {
--- a/src/components/projects/view/application-hour-tile.svelte
+++ b/src/components/projects/view/application-hour-tile.svelte
@@ -32,7 +32,7 @@
     try {
       const thisReport = report;
       application.reports = application.reports.filter(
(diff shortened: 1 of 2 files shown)


src/components/projects/view/application-hour-tile.svelte
```diff
@@ -32,7 +32,7 @@
     try {
       const thisReport = report;
       application.reports = application.reports.filter(
-        report => thisReport.reporter_email !== account.email,
+        report => thisReport.reporter_email !== report.reporter_email,
```
aabounegm, line 35: If I understand correctly, this will delete all reports by every other reporter but the current user. I believe the intention was to delete the report that was submitted by the current user, leaving reports by other moderators intact. I believe It was correct initially.

aabounegm (dismissed):

src/components/projects/view/application-hour-tile.svelte
illright: It couldn't have been correct initially since the filter function didn't depend on the element

aabounegm: Perhaps it should be something like
```suggestion
        report => thisReport !== report,
```

illright: Doesn't really matter, since the `reporter_email` uniquely identifies the report within an application, but okay, this might be faster

aabounegm: Or even just a simple `indexOf` (or `findIndex`) combined with `splice` should do the trick

aabounegm (approved):

PR #303, opened 2020-12-06
Downgrade @rollup/plugin-node-resolve

illright:


Dockerfile.prod
```diff
@@ -7,6 +7,8 @@ WORKDIR /app
 COPY .env .eslintrc.js package.json yarn.lock rollup.config.js /app/
 RUN yarn
 
+ENV HUSKY_SKIP_INSTALL=1
```
aabounegm, line 10: This should be before the `yarn` command to take effect, as it only applies to the subsequent instrucions

aabounegm (changes requested):

aabounegm (approved):

PR #306, opened 2021-01-10
Attractions v3

illright:


static/css/_attractions-theme.scss
```diff
@@ -1,62 +1,17 @@
-$font: 'IBM Plex Sans', sans-serif;
-$header-font: 'Montserrat', sans-serif;
-
+@use "sass:color";
 $main: #007a00;
```
aabounegm, line 2: Why is this one in particular left outside?

illright: It is used in the file itself later on, for the glowing shadow

aabounegm (approved):

PR #308, opened 2021-03-01
Bump dependencies

illright:
A few things to check here:
* `sapper:prefetch`
* removing manual quotes for replaced envvars (that was stupid, I admit)
* Husky v5

I've tested the thing locally, seems to work just fine, albeit some warnings in the console on build that can be ignored


aabounegm (approved):
LGTM (didn't run)

